PR #470, opened 2020-05-28
Remove gmcs/lib/itsdb.py, use PyDelphin's tsdb

goodmami:
Since PyDelphin is a dependency anyway, we might as well benefit from
the more complete implementation it provides for constructing test
suites. Note that this commit results in a few changes to the
generated test suites:

- Special characters in sentences (such as `@`) are now properly
  escaped

- Dates are formatted how TSDB expects

- The author is "Grammar Matrix Customization System" instead of
  "author-name" (I changed this; it's not a side-effect)


goodmami:
The move to Git had one more problem: the empty `tsdb/skeletons/matrix/` directory in `matrix-core` was not tracked by Git, and the existing `customize_itsdb()` function assumed it would be there, resulting in a FileNotFoundError when trying to write `tsdb/skeletons/matrix/relations`. While this could be fixed by inserting a simple `os.makedirs(matrix_skeleton, exist_ok=True)`, we already have this PR ready and it solves it with `tsdb.initialize_database(matrix_skeleton, schema=schema)` (see the [docs for `tsdb.initialize_database()`](https://pydelphin.readthedocs.io/en/latest/api/delphin.tsdb.html#delphin.tsdb.initialize_database)).

@olzama would you please look this over and merge if you don't have objections?

PR #471, opened 2020-06-02
Restructure tests

goodmami:
This moves all tests under a top level `tests/` directory instead of scattered throughout the code base. It also removes outdated regression testing code, gets unit tests working with Python 3, and removes the testing commands from `matrix.py` as there's a (implementationally) simpler way to run them.


goodmami:
@emilymbender Since Olga's trying to get the tests working I wonder if we could merge this one first. This PR touches too many files to reasonably do a thorough review of the changes (of which the vast majority are just moved files), so I suggest just looking over the commit messages above, or even the PR summary at the top.

rtest.py
```diff
@@ -213,7 +213,8 @@ def add_test(args):
     renaming them to the new test's name if necessary. Create a
     skeleton using the txt suite. Copy the skeleton to home/gold.
     Create a grammar using the current customization system, process
-    the profile. TODO: Add files to SVN.
+    the profile. It is the developer's responsibility to add these
```
olzama, line 216: Yes! Much better than relying on some extra code to add files silently to version control (there was much confusion/issues previously, I think), and can use Sourcetree etc.

olzama:
Perhaps one day we can add a --cleanup option to rtest which deletes all the logs and all the grammars (possibly also all current profiles?). It will be convenient for users who rely on GUIs to load grammars, open logs and such.

rtest.py
olzama: Actually, right after I said it it turned out that TJ (I think) recently committed two incomplete tests (he forgot the gold). So, not doing this automatically is also problematic.

goodmami: > Actually, right after I said it it turned out that TJ (I think) recently committed two incomplete tests (he forgot the gold). So, not doing this automatically is also problematic.

I still don't think it's appropriate for the code to mess with its own version control mechanisms. Rather, I plan to write some actions that check if a push is missing the files. You can already see this with `python rtest.py --list --verbose` and look for the red "no"s, but I'll make it more easily discoverable.

PR #476, opened 2020-07-05
Add missing adj-yes-no gold profiles

dantiston:
Adding missing gold profiles from r43185


dantiston:
@goodmami and @olzama

goodmami:
Looks good. I tested and now all adj-* tests pass.

Issue #20, opened 2020-05-19
question code broken (with addition of main-verb hierarchy option)


goodmami:
**Migrated from Trac:**

> - Original: <http://lemur.ling.washington.edu/trac/matrix/ticket/29>
> - Reporter: lpoulson
> - Created: 11/06/08 10:23:18
> - Updated: 10/28/11 16:07:40
> - Keywords: question
> - Notes:
>   - There may be formatting problems from the conversion
>   - Comments on tickets are *not* migrated! See the original issue.

Question code seems to be broken due to the addition of the main-verb hierarchy added when there are auxiliaries.

PR #477, opened 2020-07-10
Updated clausalmod tests, fix direct-inverse.

olzama:
I verified that all "failures" in the clausalmod tests are of the sort where previously a grammatical item was considered "not parsed" by the gold because there was no gold. In reality, the system assigns a structure to such items. I did not closely inspect the semantics of each sentence though, so in principle I could have made a mistake and committed something broken as gold (broken as a result of more recent changes). I would take me a long time to do that, and it seems likely that my wh-questions work did not affect the clausal modifiers (no other clausal modifier tests ever broke). So I think it is relatively safe to just update the tests and start treating them as correct until something proves us wrong?


olzama:
So in this branch of mine, I do not have TJ's two most recent tests; I'd removed them because they were missing gold profiles. He has since re-added them with gold profiles, and I am not sure what will happen if this pull request is merged into the trunk? @goodmami @dantiston

olzama:
Another question @goodmami , if I want to start merging my library at this point but I want to keep that separate from this pull request which has to do with the clausalmod tests, should I just start another branch from this one? This way I will have the clausalmod tests running as I think they should but the new work will not interfere with this pull request while we are still discussing it?

olzama:
Hmm OK I think I merged TJ's tests from the trunk into my branch. Not that I see that reflected in the information about the pull request though...

olzama:
Also, I tried resolving a very simple conflict in gitignore via the web interface but it is still showing me a conflict message here and furthermore now it says the conflict is too complex to be resolved here. Which is a bit puzzling.

olzama:
For the direct-inverse bug, I suggest simply casting ChoicesList to list. I realize that ChoicesList has its length method overridden ultimately, however this fix helps all the tests which were failing because of this.

goodmami:
> So I think it is relatively safe to just update the tests and start treating them as correct until something proves us wrong?

This strategy doesn't exactly fill one with confidence, but I can sympathize with your position.

> should I just start another branch from this one?

Yes, probably. In general, we should start a new branch for every changeset. Branches should be short-lived and merged (or discarded) before they get too big, because merging a very divergent branch is a lot of work.

> For the direct-inverse bug, I suggest simply casting ChoicesList to list. I realize that ChoicesList has its length method overridden ultimately, however this fix helps all the tests which were failing because of this.

Good, but I think we need to verify that doing so doesn't then misalign things (e.g., if casting to a list fills the gaps by shifting everything over).

Also, I'm not looking too closely at your committed files because this PR is still a draft and I expect that things may change. Click "Ready for review" when you want me to look at the files.

(although i did notice that clausalmods-german has some kind of comment inserted into its txt-suite)

olzama:
> Good, but I think we need to verify that doing so doesn't then misalign things (e.g., if casting to a list fills the gaps by shifting everything over).

Do you have.a suggestion on how to do this? Because the length is overwritten, I can't easily inspect ChoicesList structures in the debugger. I have to manually count and just ask the debugger for an n+x item, etc. 

Would this involve an additional pseudolanguage which deliberately uses random indexing, or something like that?

olzama:
> (although i did notice that clausalmods-german has some kind of comment inserted into its txt-suite)

yeah, no, don't care :) not fixing other people's test suites :) (I would, but there is no time right now :( ).

olzama:
> This strategy doesn't exactly fill one with confidence, but I can sympathize with your position.

Oh, I did verify, against the clausalmods paper, that the number of grammatical items is correct, and the numbers of the parsed items correspond to grammatical items in the test suites :).

goodmami:
> Do you have.a suggestion on how to do this? 

I was just thinking of walking through the failing test with some print/debug statements, understanding where it went wrong, then confirming that the proposed fix is fixing it and not giving it some unintended value.

> yeah, no, don't care :) not fixing other people's test suites :) (I would, but there is no time right now :( ).

I mean, it looks like you put it in there, probably by accident. Specifically in this commit:

https://github.com/delph-in/matrix/pull/477/commits/8181d07e0ed1af6917c694a996b7cce8b3e4caa9#diff-3ec234c1221040153be4c7ff63655dd4R12

> Oh, I did verify, against the clausalmods paper, that the number of grammatical items is correct, and the numbers of the parsed items correspond to grammatical items in the test suites :).

Alright. I trust you if you feel that's sufficient for present purposes.

dantiston:
Re: checking the clausalmods tests, I was experimenting a bit ago with the negation and infostr tests with trying to verify they passed by checking out an old revision, copying rtest into place, and testing. I didn’t have great results I think because of the changes to the regression test system since the gold profiles were committed, but checking out the original clausalmods commit and comparing them to these profiles should be pretty straightforward?

olzama:
> I mean, it looks like you put it in there, probably by accident. Specifically in this commit:

omg thank you Michael, now this is very embarrassing. Thanks for pointing this out. I will remove it :).

TJ, no, they were always empty, I think (the clausalmod golds). They just used to pass with the old system because of some error, I believe.

olzama:
> I was just thinking of walking through the failing test with some print/debug statements, understanding where it went wrong, then confirming that the proposed fix is fixing it and not giving it some unintended value.

 It was failing because it was trying to hash something that was no longer hashable in python3, isn't that right? And if it were doing something drastically different now, I doubt it would get to the point where all 5 tests would start passing. 

I get your point that the behavior wasn't well defined in the first place but by the same token, that wouldn't be a new issue introduced by something that has happened since, and so I would be inclined to put it to rest until it causes trouble once more. Again, would be happy to dedicate more attention to it but can't, really. My alternative would be to remove those tests, I think, if we don't feel comfortable enough to keep them as they are.

dantiston:
> TJ, no, they were always empty, I think (the clausalmod golds).

Sorry, I think I didn’t communicate my suggestion very well. What I’m suggesting is basically this:
1. Make 2020 gold profiles for clausalmods (which I think this commit has) and copy them elsewhere
1. Check out commit from when clausalmods was merged in or “done” (2018?)
1. Copy 2020 gold profiles and rtest into 2018 commit
1. Run rtest to compare 2018 current to 2020 gold

olzama:
TJ, sorry, I think it still sounds to me like you are assuming that at some point (in 2018), the clausalmod golds were not empty?

Oh, I think I see. Compare 2018 current to now current?

goodmami:
> It was failing because it was trying to hash something that was no longer hashable in python3, isn't that right? And if it were doing something drastically different now, I doubt it would get to the point where all 5 tests would start passing.

Oh right, the unhashable slice problem. Looking at the fix again, I think it should be fine. For whatever reason, the slice is used to get the dir-inv scales from some current index to the end. Simplified, the loop looks like this:

```python
scale_len = len(choices.get('scale', ''))
for i in range(1, scale_len):
    for scale in list(choices.get('scale'))[i:]:
        ...
```

The `__len__()` method of a choices object is the number of non-None items, and the `__iter__()` method (called when casting to a list) iterates on those non-None values. I'm reasonably convinced that this will do what we want (or rather, what Scott wanted, whatever that was).

dantiston:
> Oh, I think I see. Compare 2018 current to now current?

Right. The regression tests serve several purposes, but two relevant ones here are 1) storing the correct MRS and 2) ensuring the MRS doesn't change over time. We can't really do 1 per se (more than you have!), but thanks to SVN/git we could do 2.

So, usually regression tests are saying the gold is from the past and the current is from... the current, but I'm suggesting you could make the gold the current (as you have in 8181d07) and then check out the 2018 commit (maybe 6ff2d52?) and make the current the past :-)

olzama:
> So getting "current" for the clausalmod tests from a while ago will be not entirely straightforward because the tests were errorring before (due to some newline issue, I think). While I agree this would in principle be the right thing to do, I am not very inclined to spend any more time on this now. My preference will be to just hope they are correct now unless they start causing problems, or my second preference would be to simply remove them, on the grounds that we cannot easily figure out what they are supposed to do. Unfortunately I don't have time to work on them now. But 

Oh, that was just a pydelphin version issue. Separate from the profiles themselves.

I did verify that the currents from now are the same as the currents from the SVN trunk :).

Thank you for the suggested worksflow, T.J., I do think it was the right thing to do.

dantiston:
I was interested in trying this out, so I did some more investigating.

I checked out 7856945, did some rearranging (changed `rtest.py` back to looking at `gmcs/regression_tests`, back to using python2 for test running, etc.), and then ran the regression tests. As expected, some of the tests for those grammars failed because the gold profiles are all morphological failures:
```
clausalmods-basque
clausalmods-german
clausalmods-lavukaleve
clausalmods-madi
clausalmods-moseten
clausalmods-uranina
```

I then checked out this PR and copied the profiles from 7856945 in as gold profiles. All of the tests passed, except `clausalmods-moseten`:
```
$ delphin compare tests/regression/home/current/clausalmods-moseten/ tests/regression/home/gold/clausalmods-moseten/
1	<0,1,0>
2	<0,2,0>
5	<0,1,0>
6	<0,1,0>
7	<0,1,0>
8	<0,1,0>
9	<0,1,0>
13	<0,1,0>
14	<0,1,0>
15	<0,1,0>
19	<0,1,0>
20	<1,0,0>
21	<0,1,0>
25	<0,1,0>
```

Looking at the logs, it looks like 20 is failing to get a morphological analysis now (i.e. there was a regression).

Given the state of things, my suggestion would be to check in the gold profiles for all of these *except* remove item 20 from `clausalmods-moseten`. I can make the commit for this if you're interested, @olzama, but I've not tried collaborating on pull requests before. I suppose I could just make the commit and it should work? And then you can pull to get it?

olzama:
clausalmods-moseten passes for me (and there is a parse for item 20).
I am comparing the latest revision of the SVN trunk with what I am getting currently in git trunk. So I am not even looking at my own changes yet. But in any case, I don't think one of the older SVN commits is relevant; I think the most recent SVN trunk commit is what we should worry about?

If there was an actual interesting regression, like one particular item failing, then that should be investigated, I think (whether I will do that now or not is a different issue but I might, so long as it is something sensible rather than something I can't really interpret). But like I said, I am not seeing this regression between what we have now and the latest SVN trunk.

As for the notes about missing lexical entries in the log, those I think have always been there (we have lots of tests like these, too). So the may be unrelated to why you see item 20 failing; it passes for me and I also am seeing notes about missing lexical entries; those must have to do with other items.

olzama:
Oh wait no, you are right, it is Kristen's initial commit that's probably relevant if we wanted to "fix the test". Let me take a look as well. If it's something easy, maybe I will fix it.

olzama:
So it looks to me like this item 20 is not parsed by Kristen's original system. It is parsed by our system. And, it is marked as grammatical in the test suite, so it looks like an improvement. Maybe Kristen fixed it later, I don't know. The MRS looks good to me, the sentence means "I gave money away while moving" or something like that. So, there is a healthy looking clausal modifier there.

goodmami:
Can we ask Kristen to comment?

And the newline issue is mainly that txt-suite files didn't end in a newline, so the last item was getting skipped with the old system that used the `read` command in Bash:

```console
$ python3 -c 'print(repr(open("foo.txt").read()))'  # 2 lines in file but last line doesn't end with \n
'foo\nbar'
$ while read line; do echo "$line"; done < foo.txt  # read will miss the last line
foo
$ python3 -c 'print(repr(open("foo.txt").readlines()))'  # Python will read it
['foo\n', 'bar']
```

So I don't think that's causing problems anymore. Some files you've committed in this PR also do not terminate with `\n`, but we can clean those up later.

.gitignore
```diff
@@ -9,3 +9,15 @@ __pycache__/
...
+update
+score
+set
+rule
+phenomenon
+parameter
+output
+item-phenomenon
+fold
+analysis
```
goodmami: Aside from the missing newline after `analysis`, these are probably fine. They ignore files named with these patterns in *any* subdirectory. If we want to restrict to the directory mentioned in the comment, we could do `tests/regression/home/gold/*/update`, etc.

tests/regression/txt-suites/clausalmods-german
```diff
@@ -7,4 +7,4 @@ wir wohnten wie ich dir gesagt habe
 *wir wohnten ich dir gesagt habe wie
 *wir wohnten ich dir gesagt habe
 *wohnten wir wie ich dir gesagt habe
-*wir wohnten wie ich dir gesagt
+*wir wohnten wie ich dir gesagt
```
goodmami: Ideally this would end in a newline, even if the current system can handle it without. Fix with something like this:

```console
$ echo >> tests/regression/txt-suites/clausalmods-german
```

goodmami (approved):
Most changes are updated gold files which I did not check closely, but the rest looks pretty good. I approve these changes as-is, but see my other comments for some small suggestions.

olzama:
Thank you Mike, I will address your suggestions! 
As for asking the authors of the old tests to comment: we could but I wouldn't hold my breath. I mean, I can't even get Sanghoun to comment, and for all I know he might actually still be working on information structure (whereas most other people have industry jobs right now, so, are even less likely to pay attention). This is not to criticize the authors at all, obviously they have no obligation with respect to this. This is just to say in general it is hard to rely on this as a resource.

But let's try since I actually spoke with Kristen recently about this and she did offer to look if needed: @kphowell of you are seeing this, could you maybe take a look at item #20 in clausalmods-moseten and confirm that is should have an MRS meaning something like "I gave the money again while moving here"? (Sorry, no option to post pictures here, looks like).

tests/regression/txt-suites/clausalmods-german
olzama: Added a newline (though I don't actually know why they are needed).

goodmami: @olzama in text files, each line, including the last one, is supposed to be terminated with `\n`, so the POSIX-compliant `read` command just ignores the final sentence (as you can see in my comment [above](https://github.com/delph-in/matrix/pull/477#issuecomment-657167692)). The result was our old regression testing system created skeletons that did not include the final sentence.

olzama: That seems brittle! And is this (that any text file should end in \n) considered the expected style in python specifically, or?.. But anyway, I will try to remember this.

olzama: https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline :)

goodmami: Looks like you answered your own question :) But about the Python thing, my guess is that because Python is multi-platform, it also accommodates Windows conventions.

kphowell:
I think that's about the right meaning. The _-tom_ suffix should add a simultaneous meaning to the verb _wen_, and _wenchhishsitom_ should be the subordinated verb. Here is the (constructed by me, not vetted by a native speaker) segmentation and gloss:
yae wen-chhi-sh-si'-tom tye-baj-te kerecha
1SG  move-DC-DS-FS.NF-COM give-again-3M.O money

olzama:
Thank you so much, Kristen!

PR #478, opened 2020-07-16
Delete many unused files

goodmami:
This removes many apparently unused files:

Related to random profiles:
 - [x] gmcs/profiles.py
 - [x] gmcs/randgram.py
 - [x] gmcs/sql_profiles/ (and all files under it)

Unused library (perhaps superseded by another)
 - [x] gmcs/linglib/word_order_v2_vcluster.py

Other:
 - [x] gmcs/datestamp
 - [x] gmcs/def_check.py
 - [x] lisp/matrix-regression-tests.lsp

In addition, the following could perhaps be removed but are not included in the first commit:

- [x] lisp/matrix-powerset.lsp (this could probably be replaced with the Python equivalent if it is necessary?)
- [x] gmmt (a good candidate to move to a separate repository)
- [x] matrix-types (maybe not "unused" but it is created automatically by the install script)
- [x] gmcs/local_testing (content could be moved to a CONTRIBUTING.md file)

@emilymbender What do you think about the 4 candidates for removal above? Note that I (separately) created the [final-svn](https://github.com/delph-in/matrix/releases/tag/final-svn) tag which points to the state of the repository before the Python 3 conversion and move to Git, in case we want to recover anything later.

*edit*: removed the 4 things above after Emily approved


emilymbender:
gmmt is of historical interest only at this point, I think. I agree that matrix-powerset.lsp can be recreated if need be in python (and we haven't needed it in a long time; the need would arise if/when we want to posit an additional leaf head type that needs all of the cross-classified disjunctive supertypes).

If matrix-types is generated automatically, I don't see why it needs to be stored. And yes, local_testing looks like documentation. (Is it even current, I wonder?)

goodmami:
Done x4

I put the content of local_testing in a new CONTRIBUTING.md file, since we planned on creating one anyway. I did not verify that the instructions worked, though.

For gmmt, we can get it back from the final-svn tag, perhaps to split it into a new repository with `git subtree` (instructions [here](https://stackoverflow.com/questions/359424/detach-move-subdirectory-into-separate-git-repository)). Otherwise it's gone.

We might collect the functionality of matrix-powerset into `matrix.py` or perhaps create a `scripts/` subdirectory for these things. Python would be easier for future maintainers, I think.

goodmami:
I'm going to take Emily's comments as approval and merge this. Most of these removed things are not controversial and any that might be can be easily recovered.

PR #480, opened 2020-07-20
Use own progress bar instead of PyDelphin's

### LINKED ISSUE #479: Rtest cosmetics
I wonder if it makes sense to make the "progress bar" that rtest currently prints out, shorter? As it is, it makes the output longer than the default size of the terminal on 13'' monitors (at least on my 13" monitor). Of course it is possible to resize the terminal but that's not really the most convenient thing to have to do often (somehow).

goodmami:
Using the latest version (1.4.0) of PyDelphin, we can disable its progress bar and use our own. With this, the progress looks like this:

Screenshot from 2020-07-20 18-56-36

Note how the test name is abbreviated (`...`) when it gets too long. You shouldn't have any wrapping problems unless you resize while it's working or have a window with < 30 (or so) columns.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,1 @@
-pydelphin==1.3.0
+pydelphin==1.4.0
--- a/rtest.py
+++ b/rtest.py
@@ -5,6 +5,7 @@
 Grammar Matrix Regression Testing
 """
 
+from pathlib import Path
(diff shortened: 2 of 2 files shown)


olzama (approved):

olzama:
Thanks!!

PR #482, opened 2020-07-27
Replaced diff-lists with append lists through out; updated multiple gold MRS with ICONS.

olzama:
This one may be a bit hard to review because of how many files are involved. This version replaces all diff-lists with append-lists. A couple of bugs were exposed as a result where ICONS used to be underspecified (in some auxiliary lexical items, and in some clausal modifier phrasal rules). As a result, we got ICONS information in a lot of places where it was previously missing. I reviewed all such items manually (although somewhat quickly) and updated the gold, concluding the additional ICONS is good.
 
One change I had to do was in tdl.py; please take a look. I am not entirely confident of what I did here but it seemed necessary to do something like that because previously two lists were not merged correctly (see here: https://delphinqa.ling.washington.edu/t/replacing-rels-with-rels-list-possible-issues-with-tdl-parser/401).

The remaining failing tests are the same old information structure tests:

infostr-foc-affix-after-noun 
infostr-foc-affix-obj 
infostr-foc-affix-obj-verb  
infostr-foc-affix-subj 
infostr-foc-affix-subj-obj
infostr-foc-affix-subj-obj-hier
infostr-foc-affix-subj-obj-verb 
infostr-foc-affix-subj-verb 

These fail because we aren't getting ICONS on them append-lists do not have anything to do with this (they were failing before).

Also, I do plan to get rid of 0-alist etc., types. But I think this can be done as a separate change, perhaps. The purpose of this one is to get the system running as expected.

I wanted to request more reviewers, but can't figure out how to do that?.. It only shows me a few people as available reviewers.


goodmami:
There's not much point in hand-reviewing all the changed profile data, so if we ignore that there's not *too* much going on here. I will look at the Python code and leave the TDL to @guyemerson (assuming he agrees to review).

> I wanted to request more reviewers, but can't figure out how to do that?.. It only shows me a few people as available reviewers.

I think it's best to start with one or two only, then request additional reviewers only where it makes sense. Each reviewer request is a demand on someone's time. Also, it helps to specify in the comment what is expected of each reviewer so they don't get scared away by a huge PR.

gmcs/linglib/information_structure.py
```diff
@@ -542,7 +542,7 @@ def customize_information_structure_pos_once(mylang, ch, rules, infostr_type, in
                 #         [ SYNSEM.R-PERIPH +,
                 #           HEAD-DTR.SYNSEM [ R-PERIPH +,
                 #                             LOCAL.CONT.HOOK.ICONS-KEY $,
-                #                             NON-LOCAL.SLASH 0-dlist ] ]."""
+                #                             NON-LOCAL.SLASH 0-alist ] ]."""
                 #ph_types['subj-head-phrase'] = tdl.replace('$', infostr)
                 #ph_rules['subj-head-phrase'] = 'subj-head'
```
goodmami: Instead of changing `0-dlist` to `0-alist` in commented-out code, why not just remove the code?

gmcs/linglib/information_structure.py
```diff
@@ -564,12 +564,12 @@ def customize_information_structure_pos_once(mylang, ch, rules, infostr_type, in
...
```
goodmami, line 569: Ditto

gmcs/tdl.py
```diff
@@ -706,28 +730,37 @@ def TDLmerge(e1, e2):
...
-                        TDLmergeable(e1.child[i], e2.child[i]):
+                        TDLmergeable(e1.child[i], e2.child[i]) and len(e2.child) > 0:
                     e0.add(TDLmerge(e1.child[i], e2.child[i]))
                 else:
-                    if i < len(e1.child):
+                    if i < len(e1.child) and len(e1.child) > 0:
                         e0.add(copy.copy(e1.child[i]))
-                    if i < len(e2.child):
+                    if i < len(e2.child) and len(e2.child) > 0:
                         e0.add(copy.copy(e2.child[i]))
```
goodmami, line 721: This whole for-loop looks like it could be simplified using [zip_longest()](https://docs.python.org/3/library/itertools.html#itertools.zip_longest), e.g.:

(not tested)

```python
if e1.ordered():
    for c1, c2 in zip_longest(e1.child, e2.child):
        if c1 and c2 and TDLmergeable(c1, c2):
            e0.add(TDLmerge(c1, c2))
        else:
            if c1:
                e0.add(copy.copy(c1))
            if c2:
                e0.add(copy.copy(c2))
```

This is not only easier to read, but uses fewer comparisons and list subscripts. But only if it works :)

gmcs/tdl.py
```diff
@@ -706,28 +730,37 @@ def TDLmerge(e1, e2):
...
                 for c0 in e0.child:
                     if TDLmergeable(c0, c):
                         e0.child.remove(c0)
                         e0.add(TDLmerge(c0, c))
                         handled = True
                         break
+                    else:
+                        # Sometimes need to merge the end of list with a longer list:
+                        if isinstance(c0, TDLelem_type) and c0.type == 'null' \
+                                and isinstance(c, TDLelem_feat) and c.is_list() and not c.empty_list:
```
goodmami: Instead of `\`, you can group the conditions in parentheses. Also, the entire `else` block is a nested `if` block, so you could combine them into an `elif`:

```python
elif (isinstance(c0, TDLelem_type) and c0.type == 'null'
          and isinstance(c, TDLelem_feat) and c.is_list() and not c.empty_list):
    [...]
```

matrix-core/ace/config.tdl
```diff
@@ -27,7 +27,7 @@ handle-type               := h.
 list-type                 := list.
 cons-type                 := cons.
 null-type                 := null.
-diff-list-type            := diff-list.
+diff-list-type            := append-list.
```
goodmami: Doesn't this mean that `<! ... !>` syntax results in a `[ LIST...., LAST....]` feature structure with the type `append-list`? I haven't been following the full conversation so I'm surprised that this would help with things.

goodmami (changes requested):
I'm not familiar enough with append-lists to comment on the TDL changes, but I had some coding suggestions for the python files. I'll 'approve' when those suggestions have been considered, even if we decide against them for some reason.

gmcs/linglib/adnominal_possession.py
```diff
@@ -377,15 +377,15 @@ def customize_poss_rules(strat,mylang,ch,rules,hierarchies):
...
+                                                           '+POSSESSUM_EXIST_REL+' >,\
+                                                   HCONS.LIST < qeq & [ HARG #harg, LARG #lbl]  > ] ].')
 
         elif strat.get('mod-spec')=='mod':
             mylang.add(phrase_rule+' := [ SYNSEM.LOCAL.CAT.VAL.SPR #spr,\
                                           HEAD-DTR.SYNSEM.LOCAL.CAT [ VAL.SPR #spr & <[ ]>],\
-                                          C-CONT [ RELS <! '+POSS_REL+' !>,\
-                                                   HCONS <! !> ] ].')
+                                          C-CONT [ RELS.LIST < '+POSS_REL+' >,\
+                                                   HCONS.LIST < > ] ].')
```
dantiston, line 388: 0-alist?

gmcs/linglib/information_structure.py
olzama: I suppose!!

matrix-core/ace/config.tdl
olzama: So, this one, I don't know why it is needed but I think Woodley said it was. I will try to dig something up.

gmcs/linglib/information_structure.py
olzama: Actually, no, I don't think I want to remove Sanghoun's commented out code just yet because there are places in his library which are unfinished which I think may guide further development. So I'd leave it.

gmcs/tdl.py
olzama: So there is more code around this which uses the same style (explicit call to len); do you think it should be part of the same PR to change all of that?

gmcs/linglib/clausalmods.py
```diff
@@ -566,11 +568,11 @@ def add_morphological_subord_rel(mylang, cms, ch, rules):
...
           						COORD - ]],\
-              C-CONT [ RELS <! [ ARG1 #mch,\
-          		     ARG2 #scl ] !>,\
-          	     HCONS <! qeq &\
+              C-CONT [ RELS.LIST < [ ARG1 #mch,\
+          		     ARG2 #scl ] >,\
+          	     HCONS.LIST < qeq &\
           		    [ HARG #mch,\
-          		      LARG #mcl ] !>,\
+          		      LARG #mcl ] >, ICONS 0-alist,\
```
dantiston: New line?

gmcs/tdl.py
olzama: Addressed this one as suggested.

gmcs/tdl.py
```diff
@@ -475,6 +475,30 @@ def ordered(self):
...
+class TDLelem_list(TDLelem_feat):
+    def __init__(self):
+        self.child = []
+
+    def write(self):
+        if debug_write:
+            TDLwrite('list\n')
+
+        TDLwrite('< ')
+        for ch in self.child[0:1]:
```
dantiston: `self.child[0].write()`?

matrix-core/ace/config.tdl
olzama: @emilymbender do you remember why this change was needed? I cannot find anything in my email :/

matrix-core/matrix.tdl
```diff
@@ -1377,66 +1440,46 @@ head-valence-phrase := head-nexus-phrase &
...
+	     HCONS #h1,
+	     ICONS #i1 ],
     ARGS < sign & [ SYNSEM.LOCAL local &
-                 [ CONT [ RELS [ LIST #first,
-                                 LAST #middle1 ],
-                      HCONS [ LIST #scfirst,
-                              LAST #scmiddle1 ],
-                      ICONS [ LIST #icfirst,
-                              LAST #icmiddle1 ] ] ] ],
+				 [ CONT [ RELS #r2,
```
dantiston, line 1479: These sections look too indented (but I’m on a mobile browser, so maybe the display is bad?)

matrix-core/matrix.tdl
```diff
@@ -1642,14 +1669,13 @@ clause := phrasal &
...
-  [ SYNSEM.NON-LOCAL.REL 0-dlist,
+  [ SYNSEM.NON-LOCAL.REL 0-alist,
     HEAD-DTR.SYNSEM [ LOCAL.CONT.HOOK [ INDEX #index,
 					ICONS-KEY.IARG1 #index,
 					CLAUSE-KEY #index ],
-		      NON-LOCAL [ QUE 0-dlist,
-				  REL 0-dlist ] ],
-    C-CONT [ RELS <! !>,
-	     HCONS <! !> ] ].
+		      NON-LOCAL [ REL 0-alist, QUE 0-alist ] ],
```
dantiston: New line?

matrix-core/matrix.tdl
```diff
@@ -1832,15 +1853,15 @@ basic-head-subj-nmc-phrase := head-valence-phrase & basic-binary-headed-phrase &
...
+         NON-LOCAL.SLASH append-list & [ LIST < #slash > ] ],
     NON-HEAD-DTR.SYNSEM #synsem & canonical-synsem &
        [ LOCAL [ CAT [ VAL [ SUBJ olist,
                              COMPS olist,
                              SPR olist ] ] ],
-         NON-LOCAL [ SLASH 0-dlist & [ LIST < > ],
-                     REL 0-dlist,
-                     QUE 0-dlist ] ],
-    C-CONT [ RELS <! !>, HCONS <! !>, ICONS <! !> ] ].
+         NON-LOCAL [ SLASH 0-alist,
```
dantiston: This constraint was weird before - isn’t that a 1-dlist?

gmcs/linglib/information_structure.py
goodmami: Then I think these blocks should have some comment describing what we want to do with them. E.g., do they pertain to some specific part of his thesis? Are they related to some matrix Bug? But this is all outside the scope of this PR.

gmcs/tdl.py
goodmami: It would be good to use a consistent style if this pattern is applicable to those cases. But since those changes wouldn't be related to replacing diff-lists with append-lists, they could also be relegated to some (near-)future PR. Maybe just make a new issue for code simplification/cleanup referring to one of these comments (hint: under the `...` menu of the comment, there's a "Reference in new issue" button).

tests/regression/home/gold/clausalmods-basque/result
```diff
@@ -1,15 +1,15 @@
...
...("N" ("afaltzerat")))) ("V" ("V" ("V" ("joango"))) ("V" ("V" ("V" ("gara"))))))))@[ LTOP: h0 INDEX: e2 [ e SF: prop-or-ques E.TENSE: tense E.ASPECT: aspect E.MOOD: mood ] RELS: < [ "_when_subord_rel"<-1:-1> LBL: h4 ARG0: i5 ARG1: h6 ARG2: h7 ]  [ "exist_q_rel"<-1:-1> LBL: h8 ARG0: x9 [ x SPECI: bool COG-ST: cog-st PNG.PER: person PNG.NUM: number ] RSTR: h10 BODY: h11 ]  [ "_job_n_rel"<-1:-1> LBL: h12 ARG0: x9 ]  [ "_finish_v_rel"<-1:-1> LBL: h13 ARG0: e14 [ e SF: prop-or-ques E.TENSE: tense E.ASPECT: imperfective E.MOOD: mood ] ARG1: x15 [ x SPECI: bool COG-ST: in-foc PNG.PER: 2nd PNG.NUM: singular ] ARG2: x9 ]  [ "exist_q_rel"<-1:-1> LBL: h16 ARG0: x17 [ x SPECI: bool COG-ST: cog-st PNG.PER: person PNG.NUM: number ] RSTR: h18 BODY: h19 ]  [ "_dinner_n_rel"<-1:-1> LBL: h20 ARG0: x17 ]  [ "_go_v_rel"<-1:-1> LBL: h21 ARG0: e2 ARG1: x22 [ x SPECI: bool COG-ST: in-foc PNG.PER: 1st PNG.NUM: plural ] ARG2: x17 ] > HCONS: < h0 qeq h1 h6 qeq h21 h7 qeq h13 h10 qeq h12 h18 qeq h20 > ICONS: < e23 [ e SF: iforce E.TENSE: tense E.ASPECT: aspect E.MOOD: mood ] non-focus x15 e24 [ e SF: iforce E.TENSE: tense E.ASPECT: aspect E.MOOD: mood ] non-focus x22 > ]@((:ascore . 0.0) (:probability . 1.0))
```
dantiston, line 1: It seems concerning that these ICONS have relations between an event not referenced anywhere else in the MRS and another event.

dantiston:
Looks pretty good to me. Glad to see the simpler syntax!

gmcs/tdl.py
goodmami: I also thought this bit of code looked odd, but do note that `self.child[0].write()` would raise an IndexError if the `child` list is empty.

matrix-core/ace/config.tdl
guyemerson: I think this is necessary for generation, because ACE initialises the `CONT` lists using this type.  Unifying `diff-list` and `append-list` will fail.  (And allowing a common subtype might lead to weird bugs.)

gmcs/linglib/adnominal_possession.py
guyemerson: After discussion with Olga, we think it could be a good idea to remove the 0-alist type and related types.  They don't make the code more succinct or more readable, but they can cause issues if you want to create other subtypes (which can be useful in other cases).

So I would suggest keeping this, and changing other instances of `HCONS 0-alist` to `HCONS.LIST <>`.

matrix-core/matrix.tdl
```diff
@@ -178,15 +178,15 @@ expressed-non-canonical := non-canonical & expressed-synsem.
 
 gap := expressed-non-canonical &
   [ LOCAL #local,
-    NON-LOCAL [ REL 0-dlist,
-		QUE 0-dlist,
-		SLASH 1-dlist &
+    NON-LOCAL [ REL 0-alist,
+		QUE 0-alist,
+		SLASH append-list &
```
guyemerson: `SLASH append-list` is redundant because that is already the appropriate type for `SLASH`.  The old version was also specifying `1-dlist`, but here we can simply have `SLASH.LIST < #local >`.

I would recommend removing all of the redundant `SLASH append-list` constraints.  This is not a critical problem (after all, it evaluates to the same feature structure), and so this could also be done later, perhaps along with removing the 0-alist type.

guyemerson:
The appends themselves look good.  (With append-lists, it should be much harder now to write an incorrect append!)

In some cases, the order of the append changed.  This doesn't matter, except for bugs due to underspecified `CONT` lists in lexical items or underspecified `C-CONT` lists in rules.  (I haven't looked to see if some such lists are still underspecified, and I haven't looked at the test profiles.)

matrix-core/ace/config.tdl
emilymbender: If this is something that ACE is looking for, is it in fact something we should be changing in a config file instead?

guyemerson: Oh, I thought this *was* the config file.

emilymbender: Ah, indeed. (Sorry, I'm still really struggling with the github interface.) Then this makes sense: it's telling ACE what to treat as a "diff-list" for its internal purposes.

gmcs/tdl.py
dantiston: I always forget that Python’s slices don’t raise IndexErrors and I find that odd. It seems `if self.child: self.child[0].write()` would more accurately represent the intention, but maybe that’s beyond the scope of this PR.

gmcs/tdl.py
olzama: I created the issue.

gmcs/linglib/adnominal_possession.py
olzama: I will replace all 0-alists with LIST < > but not now. This PR is to make sure the lists are working the same way the diff-lists were (or better).

gmcs/linglib/clausalmods.py
olzama: OK :)

gmcs/tdl.py
olzama: I actually don't remember now why I wrote this class. I am not using it anywhere. I thought I needed it though... But I think I should just delete it for now... Or leave it commented out as we have sometimes been doing for unfinished work, so that it can remind us of what we were about to do... But it isn't reminding me of anything now, so...

matrix-core/matrix.tdl
olzama: There is no consistent indentation in the Matrix (unfortunately). I am not going to worry about this now. I usually try to follow what the type is already doing.

matrix-core/matrix.tdl
olzama: OK :).

matrix-core/matrix.tdl
olzama: No, it was a 0-dlist.

tests/regression/home/gold/clausalmods-basque/result
olzama: I will look!

olzama: I will open an issue for this one. I don't immediately know how to fix this, and it is unrelated to appends.

olzama: Thanks for pointing this out, TJ!

olzama:
A

olzama:
Btw, who should be resolving conversations, what policy would make sense here?

matrix-core/matrix.tdl
olzama: Thank you, I addressed this (and cleaned up the 0-alist usage in matrix.tdl; they still are present in some libraries for now).

matrix-core/matrix.tdl
dantiston: Sorry, I was wrong bringing up `1-dlist`. The main issue I wanted to address was the original constraint was `SLASH 0-dlist & [ LIST < > ]` and you've replaced it with `0-alist`. But, I take it from your other comment that `0-alist` and `LIST < >` are equivalent, so I understand now 🙂

olzama: Yeah, I don't think that 0-dlist & LIST < > was needed, either! 0-dlist implies LIST < >, I am pretty sure.

goodmami:
> Btw, who should be resolving conversations, what policy would make sense here?

I think it would be the reviewer who started the conversation, when they think it has been resolved. If they are unable (traveling, etc.), I think anybody with write-access can resolve a conversation.

olzama:
> I think it would be the reviewer who started the conversation, when they think it has been resolved.

Makes sense! I unresolved stuff :).

gmcs/tdl.py
goodmami: Ok that's fine. Let's fix them all up in another PR.

gmcs/linglib/information_structure.py
olzama: Agreed, but this is unrelated to the PR :). This comment has always been there, I just accidentally replaced the diff-list there with an append-list :). I could just revert the change :).

matrix-core/ace/config.tdl
goodmami: Ok so, as I understand, append lists are *replacing* diff lists rather than being *in addition* to them. So one could perhaps even use the `<! ... !>` syntax to describe append lists?

Also, do we need to similarly update the LKB's `*diff-list-type*` in `lkb/globals.lsp` and PET's `special-name-difflist` in `pet/flop.set`?

gmcs/linglib/information_structure.py
olzama: I went ahead and reverted this commented out part to diff-lists, perhaps there will be less confusion, more obvious that that's old code that is there for information purposes. As for adding more comments explaining what the code is for, that's separate (we can create an issue...)

goodmami: I think it's fine as-is, except for any sort of descriptive comment. If the Matrix is moving to append-lists, then the features this commented-out code supposedly implements would need to use append-lists, too.

olzama: There's lots of places like this throughout the customization system (and matrix.tdl!) Perhaps we should just be consistent (which means not touch them in this PR). I will create an issue.

goodmami: Ok that's fine.

tests/regression/home/gold/clausalmods-basque/result
olzama: to clarify: this may be related to appends generally but it is unrelated to replacing diff-lists with append-lists. We used to not get ICONS at all for these ones, now we get some ICONS, not yet correct ones, but fixing that is a task specific for the clausal modifiers library.

matrix-core/ace/config.tdl
olzama: don't know! @guyemerson @emilymbender ?

guyemerson: When I raised the question of extending TDL syntax for append lists (at the summit last year), it was pointed out that the `<! foo !>` syntax is particularly helpful because writing `[ LIST < foo . #last >, LAST #last ]` is tedious.  But with append-lists, `[ LIST < foo > ]` isn't bad.  So there's no strong motivation for special TDL syntax.

For backward compatibility, I wouldn't want to re-define the `<! !>` syntax.

I don't understand the config files, so I can't comment on that.  My previous comment was just remembering what was said in an email.

matrix-core/matrix.tdl
guyemerson: No, 0-dlist implies `[ LIST 0-1-list & #last, LAST #last ]`.  It can't have `[ LIST < > ]` or else it couldn't be appended to.  Writing `0-dlist & [ LIST < > ]` means: not only is this diff-list empty, but anything appended to it must also be empty.  I think it's unlikely that this is what was intended (and even if it *was* intended, this may lead to grammars that are difficult to maintain).

matrix-core/ace/config.tdl
goodmami: My point about the syntax is not about extending TDL for append lists, but the side effects of changing ACE's `diff-list-type`. If someone sets `diff-list-type := append-list.` *and* has `<! ... !>` in their TDL, then it would instantiate an append list with the coindexed LAST feature, I think. If so, this seems like it would cause hard-to-track-down bugs.

guyemerson: Ah, I see what you mean.  Having now read http://moin.delph-in.net/TdlRfc#Feature_interpretation_of_lists , I think I now understand this section of the config.  On closer inspection, I see that the grammar still has: `append-diff-list := append-list & diff-list.`, which is only needed as a workaround for generation.  So I'm not sure the above change to the config is actually necessary.

guyemerson: As for a longer-term solution (avoiding `append-diff-list := append-list & diff-list.`), we would need ACE and other systems to avoid non-configurable uses of diff-lists.

olzama: > I see that the grammar still has: `append-diff-list := append-list & diff-list.`

Which grammar?

goodmami:
Btw "resolving" a conversation is not necessary for merging, it's just to clean up the interface so you don't have to wade through old comments.

And I have not yet "approved" because I think the ACE config issue is still unresolved. That's my only outstanding concern.

olzama:
Right! I don't know what to do about that one, unfortunately.

matrix-core/ace/config.tdl
goodmami: > Which grammar?

I think he's referring to matrix-core, and by extension any grammar customized with it.

Can you check whether generation works *without* the `ace/config.tdl` change above? If so, maybe it's not necessary, as Guy said.

olzama: OK, sorry, this is embarrassing, but what's the simplest way to test generation? I can't figure it out (I did look in the wiki). The one way I know about involves setting up MT on the VM but I am assuming that's not necessary?

olzama: (Also, the reason I put it there is because Emily couldn't do something and Woodley suggested the change :) But yeah, I am happy to check again if I know what to check exactly).

olzama: Oh, I finally found the email thread. Here's what Emily first observed:

"A while back, Guy made a version of matrix.tdl that uses append-lists instead of diff-lists for RELS and HCONS. Olga has incorporated this into a version of the customization system that we're using in 567 this year. Alas, there's something that ace is expecting that I can't figure out how to fix via config.tdl. The attached mini-English grammar and the sentence "I sleep" provides a test case. When I try generating from the resulting MRS, I get the following error:

ubuntu@UbuntuLKB:~/Downloads/eng$ ace -g eng.dat -Tf1 | ace -g eng.dat -e
I sleep
NOTE: 1 readings, added 15 / 7 edges to chart (6 fully instantiated, 3 actives used, 3 passives used) RAM: 95k
type inference:  no glb for `append-list' and `diff-list'
  LIST appropriate to list-wrapper
  APPEND appropriate to append-list
  LAST appropriate to diff-list
ace: lexicon.c:69: lexical_dg: Assertion `wellform_result == 0' failed.
^C
Aborted (core dumped)

In order to get this far, I changed the value of diff-list-type form diff-list to append-list, but don't see what else I should be changing. Is this in fact something that can be solved via config files, or is some deeper change to ace required?"


And here's what Woodley said:

"What's going on here is that diff-lists for RELS is an assumption built into the generator.  ACE builds a diff-list of the EP specifications contained in the input MRS that align to a particular grammar sign being instantiated, and unifies it in before putting the sign into the chart.  Since ACE built a diff-list but the grammar specified an append-list, a collision occurs.

I added the following line to matrix.tdl and successfully parsed and generated a couple of simple sentences, but I don't have enough context on the implementation of the grammar to know whether this might have unintended side effects:

append-diff-list := append-list & diff-list.

Maybe Olga and Guy could comment on whether they expect the non-unifiability of diff-list and append-list to ever be something the grammar would rely on?  I can't quite imagine when that could possibly happen, especially since the supertype list-wrapper is never explicitly made the appropriate type of any feature.

Possibly a better long-term solution would be for ACE to have a config option to apply these sign specializations as append-lists instead of diff-lists, but since this patch seems to work for the moment and using append-lists for RELS is not exactly mainstream yet, I think I will hold off on that until there's a clearer need."

goodmami: Ok, thanks for the context. Woodley doesn't actually say that the config change is necessary, but the matrix.tdl one currently is. So I tried 4 variations with the `web/sample-choices/mini-english` choices file, and tried generating with:

    $ echo "the dog sleeps" | ace -g eng.dat | ace -g eng.dat -e

1. Without config change, without `append-diff-list` -> crashed
2. With config change, without `append-diff-list` -> crashed
3. Without config change, with `append-diff-list` -> succeeded
4. With config change, with `append-diff-list` -> succeeded

It seems like it makes no difference, for this test, whether the config change is made or not. Maybe we should just leave it out altogether for now?

guyemerson: Thanks for checking this.  (And sorry for my sloppy terminology!)  Yes, I think it makes sense to leave out this change.

olzama: Yeah, I see now how  I probably made this change because it seemed that Emily needed it where in fact she needed a different one. 
Great! Done.

goodmami (approved):

olzama:
@guyemerson you also don't have any objections to this being merged in at this point, I am assuming? I addressed your suggestions about cleaning up matrix.tdl. There is still some 0-alists in the customization system; I will clean them up separately.

guyemerson (approved):
Yes, looks good!

dantiston:
I'm confused about the changes for the gold profiles. I'm finding GitHub's diffs to be lacking, but it looks like for most of them, they've simply been updated from the old system to the new system? PyDelphin seems better/different about building TSDB profiles. For instance, none of the adjective tests seem to have ICONS now. As far as I can tell, the results are the same. So, why update these tests and not all of the tests? Or, were the rest of the tests somehow already updated?

olzama:
Most of the profile updates should in fact be adding ICONS.

olzama:
You are right that the adjective diffs at least do not seem to be adding ICONS (which really surprises me). I will look...

olzama:
Oh, they seem to be adding some HCONS qeqs though (which I think is good).

Screen Shot 2020-07-30 at 2 58 02 PM

dantiston:
I was just looking and it looks like for at least the adjective incorporation tests, there are additional HCONS, which look correct to me. It looks like the HCONS for the noun with the incorporated adjective missing before.

The adjective tests don't have the info-str section filled out, so they shouldn't have ICONS. I suspect the same for val-cg and maybe evidentials...

olzama:
Evidentials have ICONS I think; but regardless, all these diffs should have to do with something in the new append system causing more qeqs to show up. I did check all of them one by one but of course it is easy to miss something (as evidenced by you noticing that the ICONS were in fact disconnected). So far (apart from the disconnected ICONS which I am about to fix also) I don't think I am aware of any more bugs in the updated profiles (those additional HCONS are an improvement), but let me know if you notice more! I will go ahead and merge this in though.

dantiston:
I just finished reviewing the tests and I agree the only changes is only adding HCONS and ICONS. Oddly, in some cases, there are still missing HCONS in adjective tests, this seems to be a bug with the bare-np rule (at least for in the adjective cases). This certainly does seem to be outside the realm of this change.

For instance, this test case seems to still be missing HCONS?
```
Item:
adj1n tv n

Before:
[ TOP: h0
  INDEX: e2 [ e SF: prop-or-ques E.ASPECT: aspect E.MOOD: mood E.TENSE: tense ]
  RELS: < [ _n_n<-1:-1> LBL: h4 ARG0: x3 [ x COG-ST: cog-st PNG: png SPECI: bool ] ]
          [ _adj1_a<-1:-1> LBL: h4 ARG0: e5 [ e SF: iforce E.ASPECT: aspect E.MOOD: mood E.TENSE: tense ] ARG1: x3 ]
          [ exist_q<-1:-1> LBL: h6 ARG0: x3 RSTR: h7 BODY: h8 ]
          [ _tv_v<-1:-1> LBL: h1 ARG0: e2 ARG1: x3 ARG2: x9 [ x COG-ST: cog-st PNG: png SPECI: bool ] ]
          [ _n_n<-1:-1> LBL: h10 ARG0: x9 ]
          [ exist_q<-1:-1> LBL: h11 ARG0: x9 RSTR: h12 BODY: h13 ] >
  HCONS: < h0 qeq h1 > ]

After:
[ TOP: h0
  INDEX: e2 [ e SF: prop-or-ques E.ASPECT: aspect E.MOOD: mood E.TENSE: tense ]
  RELS: < [ exist_q<-1:-1> LBL: h4 ARG0: x3 [ x COG-ST: cog-st PNG: png SPECI: bool ] RSTR: h5 BODY: h6 ]
          [ _n_n<-1:-1> LBL: h7 ARG0: x3 ]
          [ _adj1_a<-1:-1> LBL: h7 ARG0: e8 [ e SF: iforce E.ASPECT: aspect E.MOOD: mood E.TENSE: tense ] ARG1: x3 ]
          [ _tv_v<-1:-1> LBL: h1 ARG0: e2 ARG1: x3 ARG2: x9 [ x COG-ST: cog-st PNG: png SPECI: bool ] ]
          [ exist_q<-1:-1> LBL: h10 ARG0: x9 RSTR: h11 BODY: h12 ]
          [ _n_n<-1:-1> LBL: h13 ARG0: x9 ] >
  HCONS: < h0 qeq h1 h5 qeq h7 > ]
```

There's still a missing qeq from the handle of the not modified noun's quantifier's RSTR to the noun's label, right? (So, `h11 qeq h13`).

If I recall Sanghoun's dissertation correctly, the new ICONS are associated with arg-drops, and are all `non-focus`, which seems right 🙂 

So, yay!

olzama:
Yeah, I mean, there may be any number of issues remaining, having to do with various things :). In this PR, we've replaced diff-lists with append-lists, which was its purpose. The goal was not to locate any underspecified lists, for example, although we did find and fix a couple of those, incidentally.

PR #491, opened 2020-07-31
Icons arg drop 2

### LINKED ISSUE #485: Fix ICONS in clausalmods
https://delphinqa.ling.washington.edu/t/what-should-the-icons-look-like-in-a-clausalmod-sentence/544

olzama:
OK, here's another attempt. This one I **believe** should only have real improvements; the ICONS should now be correct everywhere in the updated tests EXCEPT for valchg-lkt where they should be partially correct. I don't yet know how to fix them for items which first change valence and then go through both subject and object drop; I think this is probably a separate issue (in the valence change rather than in argument drop). I will open a separate issue but I suggest that we merge this in because (unless I missed something again) this one has only improvements and just a little of unfinished work which we will for now assume does not have to do with argument drop.


olzama:
I opened a new issue for the remaining ICONS problem (that we know of): https://github.com/delph-in/matrix/issues/492

dantiston (approved):
I think this looks good! I wonder if @sanghoun or @emilymbender can confirm?

olzama:
Merging this in because it is not clear what the ICONS even should look like for valence change! Future work (someone's).

Issue #494, opened 2020-08-04
Information structure affixes work unfinished


olzama:
After Mike write rtest, we noticed that some of the information structure tests were failing all along and we were not noticing (because the old system was not comparing ICONS).

These 8 tests are all information structure affixes tests. I investigated and was unable to locate any code that does this work (see here: https://delphinqa.ling.washington.edu/t/lost-icons-in-some-of-the-infostructure-tests/543/6?u=olzama).

I created new lexical rules for a couple of tests but I need to stop now as this isn't actually my job for the quarter. The remaining tests are:

infostr-foc-affix-after-noun
infostr-foc-affix-obj-verb
infostr-foc-affix-subj-obj 
infostr-foc-affix-subj-obj-hier
infostr-foc-affix-subj-obj-verb
infostr-foc-affix-subj-verb

They all require some analytical work figuring out what the lexical rules should look like and possibly also some work on the head-subject and head-complement rules. 

I will now merge in the two new lexical rules, tag the revision, and then remove the remaining failing 6 tests (listed above). Once someone is assigned with finishing this work, the tests will need to be restored using the tag.

olzama:
Whoever picks up this work can find the above information structure tests here:
https://github.com/delph-in/matrix/releases/tag/Last-version-with-6-failing-infostr-affix-tests

I will delete them from the trunk at this point.

goodmami:
Why not just delete the corresponding lines from `regression-test-index` so the tests are skipped (for now)?

olzama:
Oh, that's a good idea, maybe? I am trying to think whether this will be more or less confusing for someone coming in, say, a few years from now. Hmm.

goodmami:
This is what we do for the huge Sahaptin test. Also later we can change rtest.py so it shows that some tests were skipped, which could help future Matrix developers.

olzama:
Like in terms of figuring out what is going on. I think yes, you are right; having an small text file which can simply be appended to the existing index might be the easiest. Perhaps we should even have a visible file in the trunk which lists tests like that.



> This is what we do for the huge Sahaptin test.

And it did cause some confusion, as I remember :). But that's OK maybe.



> Also later we can change rtest.py so it shows that some tests were skipped, which could help future Matrix developers.

That is a great idea.

emilymbender:
I like the idea of keeping skipped regression tests in the repo and having the script just note which ones were skipped. Seems more likely they will one day get fixed :)

olzama:
[...]

PR #495, opened 2020-08-04
Added two infostr lexrules, documented unfinished work

olzama:
I would like to stop working on the information structure tests now (because I think it is not just "tests"; see below). 

I conclude that the information structure affix markers work was probably unfinished, and perhaps the MRS were created using other grammars (such as the ones with adposition markers). Then perhaps it was forgotten that the tests were only passing because the system wasn't comparing ICONS, and so we have been assuming all this time we lost the ICONS but I can only conclude that they were never there, for the affix marking tests; there is no code anywhere that would be doing anything related to affix information structure marking (there is some preliminary stuff in the morphotactics library but it is not sufficient).

I added two lexical rules from scratch, to have two of the eight tests pass: infostr-foc-affix-obj and infostr-foc-affix-subj.
They are in this PR. But now I'd like to stop and to file the remaining work as unfinished. As for the remaining failing tests, I would like to remove them but after I merge this in and tag this as the last version containing those 6 tests.

Thoughts? 

Related discussion: 
https://github.com/delph-in/matrix/issues/494
https://delphinqa.ling.washington.edu/t/lost-icons-in-some-of-the-infostructure-tests/543/6


gmcs/linglib/lexbase.py
```diff
@@ -197,6 +197,13 @@ def has_evidential(self):
                 return True
         return False
 
+    def has_infostr(self):
+        for lrt in list(self.nodes.values()):
```
dantiston: This list() idiom comes from py2to3, and you should generally avoid it. You don’t need it here.

dantiston (approved):
Sounds good to me!

gmcs/linglib/lexbase.py
olzama: OK!

olzama: I'll go ahead and clean this out everywhere I can find it, then. Should not be hard. (As usual, the reason I used it is because it was all around in the surrounding code :) ).

olzama: Done.

goodmami: It's unnecessary for iteration (`for x in something.values(): ...`) but it is still often necessary when you want the values as a list.

gmcs/linglib/lexbase.py
```diff
@@ -166,13 +166,13 @@ def roots(self):
...
         """
-        return [n for n in list(self.nodes.values()) if len(n.parents()) == 0]
+        return [n for n in self.nodes.values() if len(n.parents()) == 0]
 
     def input_span(self):
         return self.hierarchy.get_lineage(key=self.key, relation='input')
 
     def valid_inputs(self):
-        all_inps = list(self.input_span().values())
+        all_inps = self.input_span().values()
```
goodmami: Here you probably want to keep the `list()`.

goodmami (approved):
On a quick glance the code looked fine except for one thing. I didn't run the tests or inspect the outputs so I'll take your word for it.

gmcs/linglib/lexbase.py
olzama: Hmm I did run the tests. But I will put it back I guess.

goodmami: It's possible that the return value of `valid_inputs()` is only iterated over (or other operations that don't need a list), but it's better for our functions to return lists than view objects to avoid problems down the road.

dantiston: Sorry, I probably should have been more specific about the weird list thing! It’s specifically the `for x in list(d.keys())`/`for x in list(d.values())`. As @goodmami indicates, in Python 2, keys/values return a list, whereas in Python 3 they return special keys/values view objects, so just iterating over them is an unnecessary cast, but otherwise it’s probably good to keep them.

PR #496, opened 2020-08-04
Remove infostr affix tests

olzama:
Per Michael's suggestion, I removed the information structure tests which I believe were always failing, from the index. So the tests are still there but are excluded from the test run. I also added a list of currently skipped tests, so that this is all more visible. Do people think this list is needed?


olzama:
https://github.com/delph-in/matrix/issues/494 Discussion.

goodmami:
Thanks! I think it might be nice to keep one index. See #497. I can get that working fairly quickly.

If you agree, there's no need to close this PR; we can just amend it with new commits and I'll "squash" them when merging.

PR #498, opened 2020-08-04
rtest.py usability improvements

### LINKED ISSUE #497: Add better support for "skipped" regression tests
There is some need to allow known-failing tests to be skipped (see #494). We can remove the entries for the tests from `tests/regression/regression-test-index`, as is done for Sahaptin, but this index also contains comments, so deleting them would lose information. We could create a secondary index of skipped tests, or we could somehow mark tests in the index to be skipped (similar to pytest's [mark.skip](https://docs.pytest.org/en/stable/skipping.html) functionality). For the latter option, how about a `!` before the entry?

```
infostr-foc-adp-after-noun=infostr
!infostr-foc-affix-after-noun=infostr
```

Here the second test is skipped.

The other part of this issue is finding a better way to report that tests were skipped. Tests with missing files (txt-suites, skeletons, etc.) or entries in this file would be marked as incomplete, and those with `!` in the index would be reported as skipped.

goodmami:
**Edit:** I updated the title and body to broaden the scope of this PR a bit.

### Skipped Tests

There is now a notion of "skipped" tests, which come in two flavors: "disabled" tests are those annotated in `regression-test-index` with `!`, meaning they are to be skipped; and "incomplete" tests are those that are missing some required files (choices, skeleton, gold profile) or are not indexed.

- [X] add SKIP result category
- [X] add `!` prefix to regression-test-index format for disabled tests
- [X] replace `--all-tests` with `--skipped` which limits test discovery to skipped tests
- [X] print skipped status and reason with `rtest.py --list --verbose`

Some more testing is needed. In particular:

- [x] check behavior with a disabled test
- [x] ensure adding tests still works
- [x] ensure removing tests still works

Resolves #497 

### Simplified Testing Pipeline

Now that `grm.dat` files are deleted, the current pipeline makes less sense. So now:

- [x] remove `--mkprof` option; instead, create profile from skeleton during `--process`
- [x] compile and delete `grm.dat` during `--process`
- [x] run customization when `--process` is used, even if `--customize` is not

Additionally:

- [x] prevent `--mkskel` from creating non-skeleton files (edge, result, etc.)

### Other

A few other related changes to help with scripting and debugging:

- [x] `rtest.py` returns a nonzero exit code if tests fail or have errors
- [x] `--debug` turns off features that may interfere with a debugger, namely multiprocessing
- [x] `--clean` deletes customized grammars, logs, and current profiles

### DIFF
--- a/rtest.py
+++ b/rtest.py
@@ -65,6 +65,7 @@ DONE = 'DONE'
 PASS = 'PASS'
 FAIL = 'FAIL'
 ERROR = 'ERROR'
+SKIP = 'SKIP'
 
 
 def linewidth(): return min(MAX_LINE_WIDTH, shutil.get_terminal_size()[0])
@@ -83,6 +84,7 @@ REPORT_COLOR = {
     DONE: green,
     PASS: green,
     FAIL: yellow,
+    SKIP: yellow,
     ERROR: red
 }
 
@@ -133,6 +135,7 @@ def run_tests(args):
         ERROR: 0,
         FAIL: 0,
         DONE: 0,
+        SKIP: 0
     }
 
     tests = list(_discover(args))
@@ -161,7 +164,11 @@ def run_tests(args):
         width = len(str(total))  # to align the numbers on /
         print('Passed {0:{2}}/{1} tests;'.format(totals[PASS], total, width))
         print('Failed {0:{2}}/{1} tests;'.format(totals[FAIL], total, width))
-        print('Errors {0:{2}}/{1} tests.'.format(totals[ERROR], total, width))
+        print('Errors {0:{2}}/{1} tests;'.format(totals[ERROR], total, width))
+        if totals[SKIP]:
+            print('Skipped {0:{2}}/{1} tests.'
+                  ' (run rtest.py --list --skipped --verbose for more info)'
+                  .format(totals[SKIP], total, width))
 
 
 def _run_test(
@@ -170,40 +177,41 @@ def _run_test(
         mkskel=False,
         mkprof=False,
         process=False,
-        compare=False
+        compare=False,
 ) -> Tuple[str, str, pathlib.Path]:
-    name, desc, chc, dat, txt, skel, prof, gold = args
+    name, idx, chc, dat, txt, skel, prof, gold = args
     log = _unique_log_path(name)
-    result = DONE  # default if no error, failure, or comparison pass
+    result = DONE  # default if no skip, error, failure, or comparison pass
 
     with log.open(mode='at') as logf:
         _lognow('== Testing {} at {} =='
                 .format(name, datetime.datetime.now().isoformat()),
                 logf)
 
-        try:
-            if customize:
-                grm = _customize(name, chc, logf)
-                dat = _compile(name, grm, logf)
-            # mkskel if requested or if necessary
-            if mkskel or (mkskel is None and skel is None):
-                skel = _mkskel(name, txt, logf)
-            if mkprof:
-                prof = _mkprof(name, skel, logf)
-            if process:
-                _process(name, dat, prof, logf)
-            if compare:
-                passed = _compare(name, prof, gold, logf)
-                result = PASS if passed else FAIL
-
-        except Exception:
-            _lognow('\n=====', logf)
-            traceback.print_exc(file=logf)
-            result = ERROR
-
-        # delete the dat file; it's a big file that's easy to recreate
-        if dat.exists():
-            dat.unlink()
+        if _skipped(idx, chc, skel, gold):
+            result = SKIP
+        else:
+            try:
+                if customize:
+                    grm = _customize(name, chc, logf)
+                    dat = _compile(name, grm, logf)
+                # mkskel if requested or if necessary
+                if mkskel or (mkskel is None and skel is None):
+                    skel = _mkskel(name, txt, logf)
+                if mkprof:
+                    prof = _mkprof(name, skel, logf)
+                if process:
+                    _process(name, dat, prof, logf)
+                if compare:
+                    passed = _compare(name, prof, gold, logf)
+                    result = PASS if passed else FAIL
+            except Exception:
+                _lognow('\n=====', logf)
+                traceback.print_exc(file=logf)
+                result = ERROR
+            # delete the dat file; it's a big file that's easy to recreate
+            if dat.exists():
+                dat.unlink()
 
         _lognow('\nResult: ' + result, logf)
         return name, result, log
@@ -214,20 +222,33 @@ def list_tests(args, verbose=False):
     Print each test name and exit.
 
     If --verbose is used, also print the description, whether the test
-    appears in the index, and whether it has an associated txt-suite,
-    choices file, skeleton, and gold profile.
+    appears in the index, whether it has an associated txt-suite,
+    choices file, skeleton, and gold profile, and whether it is to be
+    skipped.
     """
-    for name, desc, chc, dat, txt, skel, prof, gold in _discover(args):
+    for name, idx, chc, dat, txt, skel, prof, gold in _discover(args):
 
         print(name)
 
         if verbose or args.verbosity >= 2:
-            print('    Description: ' + desc)
-            print('    Indexed:     ' + yes_or_no(desc))
-            print('    Txt-suite:   ' + yes_or_no(txt))
-            print('    Choices:     ' + yes_or_no(chc))
-            print('    Skeleton:    ' + yes_or_no(skel))
-            print('    Gold:        ' + yes_or_no(gold))
+            if idx:
+                desc = idx.get('description')
+                if idx.get('skip'):
+                    skip = green('yes')
+                elif _skipped(idx, chc, skel, gold):
+                    skip = f'{red("yes")} (missing components)'
+                else:
+                    skip = green('no')
+            else:
+                desc = None
+                skip = f'{red("yes")} (not indexed)'
+            print('    Description:', desc)
+            print('    Indexed:    ', yes_or_no(idx))
+            print('    Txt-suite:  ', yes_or_no(txt))
+            print('    Choices:    ', yes_or_no(chc))
+            print('    Skeleton:   ', yes_or_no(skel))
+            print('    Gold:       ', yes_or_no(gold))
+            print('    Skipped:    ', skip)
             print()
 
 
@@ -238,7 +259,7 @@ def update_test(args):
     tests = list(_discover(args))
     if len(tests) != 1:
         raise RegressionTestError('only 1 test may be updated at a time')
-    name, desc, chc, dat, txt, skel, prof, gold = tests[0]
+    name, idx, chc, dat, txt, skel, prof, gold = tests[0]
 
     try:
         db = tsdb.Database(prof)
@@ -258,20 +279,19 @@ def add_test(args):
     the profile. It is the developer's responsibility to add these
     files to Git afterwards.
     """
-    args.all_tests = True  # necessary for adding new tests
     tests = list(_discover(args))
     if len(tests) == 0 and len(args.test) == 1:
         name = args.test[0]
-        desc, chc, txt = None, None, None
+        idx, chc, txt = None, None, None
     elif len(tests) == 1:
-        name, desc, chc, _, txt, _, _, _ = tests[0]
+        name, idx, chc, _, txt, _, _, _ = tests[0]
     else:
         raise RegressionTestError('only 1 test may be added at a time')
 
     msg = ('File {!s} already exists; if you want to update the test, use '
            'the --update command; if you are sure the file is rogue, delete '
            'it. Otherwise, use a different name for a new test.')
-    if desc is not None:
+    if idx is not None:
         raise RegressionTestError(
             'test already exists in index: {}'.format(name))
     if chc is not None:
@@ -282,7 +302,7 @@ def add_test(args):
 
     desc = input('Test description: ')
     index = _parse_index(args.index)
-    index[name] = desc.strip()
+    index[name] = {'description': desc.strip()}
 
     # Copy files and recreate the index
     shutil.copy(str(chc), str(CHOICES_DIR / name))
@@ -338,11 +358,10 @@ def remove_test(args):
     the developer's responsibility to commit these changes to Git
     afterwards.
     '''
-    args.all_tests = True  # necessary for partial tests
     tests = list(_discover(args))
     if len(tests) > 1:
         raise RegressionTestError('only 1 test may be removed at a time')
-    name, desc, chc, dat, txt, skel, prof, gold = tests[0]
+    name, idx, chc, dat, txt, skel, prof, gold = tests[0]
 
     for obj in (chc, txt, skel, prof, gold):
         if obj is not None:
@@ -377,9 +396,8 @@ def _discover(args):
     directories of [incr tsdb()] skeletons, current profiles, and gold
     profiles, joined by a shared test name. Collect and merge these
     sources of information and yield each as a tuple of (name,
-    description, txt-suite-path, choices-path, skeleton-path,
-    current-path, gold-path). If --all-tests is used, partially
-    described tests not in the index will be yielded, too.
+    index entry, txt-suite-path, choices-path, skeleton-path,
+    current-path, gold-path).
     """
     index     = _parse_index(args.index)
     choices   = _list_files(CHOICES_DIR)
@@ -389,27 +407,36 @@ def _discover(args):
     profiles  = _list_testsuites(CURRENT_DIR)
     gold      = _list_testsuites(GOLD_DIR)
 
-    all_names = (set(index)
-                 .union(choices)
-                 .union(grammars)
-                 .union(txtsuites)
-                 .union(skeletons)
-                 .union(profiles)
-                 .union(gold))
-    if not args.all_tests:
-        all_names = filter(index.__contains__, all_names)
-    all_names = sorted(all_names)
+    all_names = sorted(
+        (set(index)
+         .union(choices)
+         .union(grammars)
+         .union(txtsuites)
+         .union(skeletons)
+         .union(profiles)
+         .union(gold))
+    )
 
     for pattern in args.test:
         for name in fnmatch.filter(all_names, pattern):
+            idx = index.get(name)
+            chc = choices.get(name)
+            skl = skeletons.get(name)
+            gld = gold.get(name)
+            if args.skipped and not _skipped(idx, chc, skl, gld):
+                continue
             yield (name,
-                   index.get(name),
-                   choices.get(name),
+                   idx,
+                   chc,
                    grammars.get(name),
                    txtsuites.get(name),
-                   skeletons.get(name),
+                   skl,
                    profiles.get(name),
-                   gold.get(name))
+                   gld)
+
+
+def _skipped(idx, chc, skel, gold):
+    return None in (idx, chc, skel, gold) or idx.get('skip')
 
 
 def _parse_index(path):
@@ -418,22 +445,28 @@ def _parse_index(path):
     for line in path.open():
         line = line.strip()
         if line:
-            name, description = _split_index_line(line)
-            index[name] = description
+            name, data = _parse_index_line(line)
+            index[name] = data
     return index
 
 
 def _recreate_index(index, path):
     """Overwrite the index with an updated one."""
     with path.open('w') as f:
-        for _name, _desc in index.items():
-            print('{}={}'.format(_name, _desc), file=f)
+        for name, data in index.items():
+            skip = '!' if data.get('skip') else ''
+            print(f'{skip}{name}={data["description"]}', file=f)
 
 
-def _split_index_line(line):
-    """Return the name and description for an index entry."""
+def _parse_index_line(line):
+    """Return the name and associated data for an index entry."""
+    data = {}
+    if line.startswith('!'):
+        data['skip'] = True
+        line = line[1:]
     name, _, description = line.partition('=')
-    return name, description
+    data['description'] = description
+    return name, data
 
 
 def _list_testsuites(dir):
@@ -449,7 +482,7 @@ def _list_files(dir):
     """Map basename to path for files in *dir*."""
     paths = {}
     for path in dir.glob('*'):
-        if path.is_file() and path.name != 'README':
+        if path.is_file() and path.name not in ('README', 'README.md'):
             paths[path.name] = path
     return paths
 
@@ -668,12 +701,12 @@ if __name__ == '__main__':
     parser.add_argument('--index',
                         metavar='PATH',
                         help='path to a test index')
-    parser.add_argument('--all-tests',
-                        action='store_true',
-                        help='don\'t exclude tests not in the index')
     parser.add_argument('-l', '--list',
                         action='store_true',
                         help='list available tests (-v, -vv for more info)')
+    parser.add_argument('--skipped',
+                        action='store_true',
+                        help='find skipped/incomplete tests, ignore others')
     parser.add_argument('-c', '--customize',
                         action='store_true',
                         help='customize and compile test grammars')
--- a/rtest.py
+++ b/rtest.py
@@ -73,6 +73,7 @@ def linewidth(): return min(MAX_LINE_WIDTH, shutil.get_terminal_size()[0])
 
 # ANSI colors; see: https://en.wikipedia.org/wiki/ANSI_escape_code#Colors
 def red(s): return color('\x1b[31m', s)
+def boldred(s): return color('\x1b[31m\x1b[1m', s)
 def green(s): return color('\x1b[32m', s)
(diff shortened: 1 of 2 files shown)


goodmami:
Ok I've added most of the things I listed. For `--debug`, I still wasn't able to get the breakpoints to work in Visual Studio Code, and I realized that either way customization happens in a separate process (it has always done this), so I'm not sure if I'm even using the debugger correctly. @olzama I would appreciate if you could confirm whether it works with or without `--debug`.

The `--clean` option is a different mode, like `--list` or `--add`. It can take a parameter to only clean certain tests (e.g., `rtest.py --clean infostr\*`), or otherwise it cleans all.

olzama:
I tried it out! The tests run somewhat faster than before although from the standard output, it looks like it only occasionally "spits out" more than one test result, and the rest of the time it appears to still go one by one, with the same speed as before. But it could of course just be the standard output that makes this impression.

Debugging in PyCharm does appear to work for me, at a first glance (even without the `--debug` option). I am still worried that it will one day not work when I most need it, or something like that :). Let us at least tag the version which is completely multiprocessing-free, in case there are issues down the road? Or is the `--debug` option exactly, 100% that?

It does not seem possible to step into either `imap` or `map` function, but I don't think that has to do with multiprocessing. So, in that case, you have to know where to put the breakpoint down the line. Not ideal but happens.
```
    if args.debug:
        results = map(run_test, tests)
        totals = _accumulate(results, total)
    else:
        with multiprocessing.Pool(PROCESSES) as pool:
            results = pool.imap(run_test, tests, chunksize=BATCH_SIZE)
            totals = _accumulate(results, total)

```

olzama:
> I still wasn't able to get the breakpoints to work in Visual Studio Code,

Actually, where did you try to put the breakpoints? Here's what I tried, for example: breakpoint at line 1142 in information_structure.py and run infostr-foc-mod-after-noun

rtest.py
```diff
@@ -65,6 +65,7 @@
 PASS = 'PASS'
 FAIL = 'FAIL'
 ERROR = 'ERROR'
+SKIP = 'SKIP'
```
olzama, line 70: Maybe a different color for SKIP than for FAIL? Otherwise easy to not notice a new FAIL?

rtest.py
```diff
@@ -366,6 +385,35 @@ def remove_test(args):
...
+            shutil.rmtree(prof)
+            deleted = True
+        # a test name may be a substring of another test name, so use a
+        # glob pattern that looks like a date to mitigate this
+        date_glob = '[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]'
+        for log in LOGS_DIR.glob(f'{name}-{date_glob}*'):
+            log.unlink()
+            deleted = True
+        if deleted:
+            print(f'cleaned files for {name}')
```
olzama: I am on the fence about printing out each deleted filename; on the one hand, this is transparent (good); on the other, with thousands of files, this will be a bit annoying. After all, you ran the command in order to get rid of them quickly. To print out several thousand filenames may take more than a minute, and at that point might as well delete them manually?

olzama (commented):
I did not retest all the functionality (`--add, --remove`, in particular). It should not be affected, right? I just had a couple cosmetic suggestions (and as for what I perceive as the "dangers" of introducing concurrency, I suppose we can tag the previous version, hope for the best (and for the `--debug` function to save us), and enjoy the speed up :).

rtest.py
goodmami: It only prints the test name for which files were deleted, not for every file. Manually deleting (typing or globbing every path, or building a `find` query, making sure not to delete the `README.md` files, etc.) them takes much longer than the time to print the test names. Personally I find it a bit reassuring to see what happened, but if it's really annoying we can hide it unless `--verbose` is used.

rtest.py
goodmami: Hmm, how about fail is red, and error is bold red?

rtest.py
olzama: > It only prints the test name for which files were deleted, not for every file.

Yeah, that's thousands, if you haven't cleared your logs for a while.



> Manually deleting (typing or globbing every path, or building a `find` query, making sure not to delete the `README.md` files, etc.) them takes much longer than the time to print the test names.

I don't think so, actually! I usually just restore the READMEs by stashing local changes. Takes a few seconds, but of course is still annoying to have to do by hand.

> Personally I find it a bit reassuring to see what happened, but if it's really annoying we can hide it unless `--verbose` is used.

This would be my preference fwiw :).

rtest.py
olzama: Btw, there should **never** be a new SKIP, right?.. Unless one puts it in by hand?

goodmami:
Thanks for trying it out.

>  looks like it only occasionally "spits out" more than one test result, and the rest of the time it appears to still go one by one, with the same speed as before

The multiprocessing method I use ensures the results are returned in "correct" order. So it shoots off 4 (or so) processes at once, but doesn't print anything until the first process finishes. If the first one is slower than the rest, then you might see a bunch printed simultaneously. If they finish in order, you'll more easily notice each one being printed. This shouldn't have an effect on the overall time; it just makes the reporting a bit less smooth.

If you have ~4 cores, you should see that `time ./rtest.py` is 2 or more times faster than `time ./rtest.py --debug`. It's not 4x faster, but it still feels like a big improvement.

> Debugging in PyCharm does appear to work for me, at a first glance (even without the `--debug` option).

You mean "even *with* the `--debug` option"? And debugging `rtest.py` worked before? I ask because it was already starting subprocesses for customization. Maybe it's the `map()` generator that's the problem.

> Let us at least tag the version which is completely multiprocessing-free, in case there are issues down the road?

As long as we can delete the tag when things work. I'd rather keep tags for big events like releases or versions associated with publications, otherwise these interesting tags get lost in the weeds.

> Or is the `--debug` option exactly, 100% that?

It's 100% multiprocessing free, but the code is not exactly the same as before. Now it seems like the `map()` thing might be the problem.

rtest.py
goodmami: Or if one creates a broken test by forgetting to commit something (choices, gold profile, etc.) or doesn't add it to the index. Recall there are "disabled" skips and "incomplete" skips.

rtest.py
goodmami: > Yeah, that's thousands, if you haven't cleared your logs for a while.

No. We have ~475 tests that run (the skipped tests don't produce logs). Like I said, only the test name is printed, not each file.

goodmami:
I missed some questions...

> I did not retest all the functionality (`--add`, `--remove`, in particular). It should not be affected, right? 

These used to set the `--all-tests` option, which is now gone. I think that was only to ensure that we didn't add an existing test name or remove a non-existing one? If so, maybe we could make `--skipped` mutually exclusive with `--add` or `--remove`.

> Actually, where did you try to put the breakpoints?

I put one on the first statement of `gmcs.customize.customize_matrix()`. I also tried removing the map and using a normal for-loop in `rtest.py`, but I still get nothing. I think there's some additional setup I need to do with vscode to get debugging working, but I'm not familiar (and it's late here, so I'll try another time).

olzama:
> > Debugging in PyCharm does appear to work for me, at a first glance (even without the `--debug` option).
> 
> You mean "even _with_ the `--debug` option"? And debugging `rtest.py` worked before? I ask because it was already starting subprocesses for customization. Maybe it's the `map()` generator that's the problem.

No, I meant "without". Before, it worked in some cases but probably not with map() etc. Like I said, that's different; that's just not a function that the PyCharm debugger can step into (like lambda functions etc.)

rtest.py
dantiston: I can’t find a reference now, but I believe there is a precedent for using blue to indicate skipped tests in Python.

olzama: > Or if one creates a broken test by forgetting to commit something (choices, gold profile, etc.) or doesn't add it to the index. Recall there are "disabled" skips and "incomplete" skips.

Hmm OK, so, they are potentially distracting, then. I am thinking from the point of view, you are running tests, and it is easy to notice something's off if you used to have all green, zero failures of any kind, and now suddenly you've got something. Having 23 skipped tests (what we have now) is then not ideal because now you have those 23 items to think about, too. Is it still 23, or did it become 24? Has it always been 24? etc. Not a clean situation when you know for sure things are fine and nothing changed... Then again, this has always been the case, I guess; the SKIP output simply makes it explicit (which is a good thing).

olzama: I guess at least SKIP gives you confidence that it is not some  difference in the MRS.

Actually, wouldn't it be better to treat disabled and incomplete as separate? The incomplete ones I think need to be either deleted or completed... They are broken.

olzama:
@curtosis Do you remember if you meant to delete or keep this test? 

valchg-jpn-rom
    Description: None
    Indexed:     no
    Txt-suite:   yes
    Choices:     yes
    Skeleton:    yes
    Gold:        yes
    Skipped:     yes (not indexed)

It looks like we have everything for it but it is not being run. Any recollection?.. Should it be indexes and skipped or should it be deleted?

rtest.py
olzama: If the number of lines printed is the number of tests maximum every time, then it is less of a problem, though I still think it makes the process slower (but still automatic) than removing everything manually via Finder (particularly removing just the logs, which is all I usually need).

rtest.py
olzama: To clarify: I think getting used to some 23 skipped tests (some of which may be broken) flashing in front of you in the output leads to just becoming insensitive to that information. So, the situation becomes similar to keeping some failed tests around which we did for a while with information structure, and it is not ideal because you sort of become accustomed to it and start ignoring this kind of information altogether (at least to a degree). So I want to have a situation where there is confidence that nothing changed between two runs, and that everything that is skipped is skipped because it was disabled by hand. The rest of them I think should be either removed or indexed and disabled, and the new ones which were added improperly should be ERRORING?

goodmami: > I believe there is a precedent for using blue to indicate skipped tests in Python.

I was following `pytest`'s convention, which uses yellow (and red for fail/error). If blue makes more sense for people, that's fine.

> Hmm OK, so, they are potentially distracting, then. [...]

I understand it may make you feel uneasy to see something not green, but these have been here for years. The change is just making them visible so we can act on them. These have been here for some time (see my thread on the matrix-dev list from about a year ago titled "Auditing the Matrix regression test system").

> I guess at least SKIP gives you confidence that it is not some difference in the MRS.

A skipped test is not a failed test, nor is it a passed test. I don't think it should grant one any confidence nor particular unease.

> Actually, wouldn't it be better to treat disabled and incomplete as separate?

I don't think so. If we make them errors, then the tests fail, and also they'd be lumped imperfectly into another category: things that raised actual errors. From the point of view of running the regression tests, the result of the test is that it was skipped, not that it passed, failed, or raised an error. If you want to see *why* something was skipped, `rtest.py --list --skipped --verbose` (or `rtest.py -lv --skipped` for short) gives you that information.

> I think getting used to some 23 skipped tests (some of which may be broken) flashing in front of you in the output leads to just becoming insensitive to that information [...]

I sympathize with that, but the point of skipped tests (instead of ignoring known failures) is that we've decided that they are not counted as failures. They're more like reminders. Ideally they would be fixed and re-enabled, but we can't make everything perfect at once.

goodmami:
@olzama I would rather leave this PR for improvements to rtest.py and actually removing tests as a separate PR. That way they can get proper discussion. It's not necessarily the case that someone incompletely removed the tests---they may have been incompletely added, and some cases might not be difficult to fix.

(*edit*: sorry for the incorrect @-mention, `@olga`!)

olzama:
Sure, I can open a separate PR for that. Although the ones I already deleted I am pretty sure about, so, perhaps we can leave them out? Or do you think it is better to roll back.

rtest.py
olzama: > If we make them errors, then the tests fail, and also they'd be lumped imperfectly into another category: things that raised actual errors.

I think I don't see how an incompletely added (or incompletely removed) test is not an actual error?

goodmami: It's a different category of error, then? The "error" tests are something like an exception raised by Python, or ACE crashing. These are just not run because some of the necessary files are missing. I guess we could force it to an actual error by trying to run the tests with missing files anyway, but I don't like that option.

goodmami:
I tried to roll back the commits but that merge commit brought them back into the history. @olzama, sorry, I should have warned you first. I will force-push again, but please do this (you will lose any unsaved work in the current branch or working directory):

    git reset --hard bd2df271a9269f314ab4a5a14dd9b8e841547806
    git pull

Then please don't push if you see conflicts or if you make a merge commit. It should apply cleanly ("fast-forward").

The commits for the removed tests are preserved in a new branch https://github.com/delph-in/matrix/tree/remove-skipped-tests, based on trunk.

rtest.py
olzama: Yes, I think it is a kind of error, different internally from ACE crashing (but ultimately not that different from Python raising an exception, which is exactly what would happen if the test were run).

> I guess we could force it to an actual error by trying to run the tests with missing files anyway, but I don't like that option.

Why not? What would be a scenario in which it makes sense to just keep a test like this in its incomplete form? I don't like that they will be lumped together with the ones which we explicitly disabled by hand. If I by accident delete some part of the test, it will become a SKIP and I won't even notice because I will be assuming that it had been disabled. 

Remembering exactly which tests were in which state in the previous run is a great skill to have and I think most people do not have it. It is easy to not notice a change in a situation like this. So, it is not that I simply want to always see all green :). I just want to avoid a pitfall here (I've been working with hundreds of tests for a long time now).

olzama:
I think it all worked out:

Screen Shot 2020-08-04 at 7 47 07 PM

goodmami:
Great. I hope you enjoy the latest commit :)

olzama:
Would you say there is now some initial overhead which perhaps leads to the time required to run, say, a single test being larger than before? (It is not a big deal for me; just wondering.)

rtest.py
olzama: Thank you! Love the new --clean option :).

goodmami:
It probably has to do with the number of inputs or the sentence length. You can run the following command to get the number of sentences and average sentence length for each test:

```bash
for d in tests/regression/skeletons/*; do
  echo $( basename "$d" ) \
       $( delphin select i-input "$d" \
        | awk '{ words += NF; sents++ } END { printf "%d\t%g", sents, words/sents }' )
done
```

rtest.py
olzama: Oh! Something just occurred to me. If you'd rather have the default --clean option be verbose (say it speaks to you conceptually, you believe such a command should by default be printing out filenames), I'd be completely happy with an additional option like --clean --suppressed or whatever. So long as I can quickly delete everything and not wait, that'll work.

goodmami: I just think that destructive actions like this shouldn't always work silently. The current solution is fine.

rtest.py
goodmami: > Why not? What would be a scenario in which it makes sense to just keep a test like this in its incomplete form? I don't like that they will be lumped together with the ones which we explicitly disabled by hand. If I by accident delete some part of the test, it will become a SKIP and I won't even notice because I will be assuming that it had been disabled.

Ok, thank you for persisting on this topic. I'm coming around to your point of view. Ultimately, yes, these are broken tests and we probably should flag them as errors if new broken tests are created somehow. One reason they are skipped now is because they had been ignored (but not explicitly skipped) for so long, and when I first created the rtest.py replacement for the old regression test system, I wanted it to replicate the previous behavior while making it easy to discover the now-visible existing problems. If we're ready to deal with all the broken tests (#499) then I'm prepared to make these proper errors.

olzama: Yes, I will deal with the broken tests (as part of https://github.com/delph-in/matrix/pull/499).

goodmami:
I tested `--add` and `--remove`. There were in fact some problems with `--add`, so those are now fixed. This resulted in a change of behavior. The `--skipped` option, as before, restricts test discovery to skipped items, but now it will force those items to be processed when running tests and not just listing them. This helped solve a problem where `--add` constructs the gold profile by processing the test suite and it was (according to the new logic) skipping these "incomplete" tests.

I did not change the behavior for incomplete tests; currently they still act like skipped tests. This is because the ability to force tests to run reveals that some incomplete tests pass as-is:

```
$ python rtest.py --skipped
Sahaptin                                                                 PASS
adnom-poss-mod-dep-aff-head-non-aff-hf                                   ERROR
adnom-poss-mod-dep-non-aff-head-non-aff-hf                               ERROR
adnom-poss-spec-head-aff-hf-block-pron                                   ERROR
adnom-poss-spec-possessor-affix-two-strat-mod-spec-var                   ERROR
adnom-poss-spec-possessor-affix-two-strat-order-var                      ERROR
adnom-poss-spec-possessor-non-affix-possessum-non-affix-bidirec          ERROR
ccomp-extra-and-same                                                     PASS
ccomp-oblig-bef-aft-ovs                                                  PASS
ccomp-oblig-svo-form                                                     PASS
ccomp-v2-vfinal-aux                                                      FAIL
clausalcomps-v2-vfinal-aux                                               ERROR
clausalmods-mod-position                                                 ERROR
neg-head-comp                                                            ERROR
neg1b-neg2fd                                                             ERROR
v2-orig                                                                  ERROR
valchg-jpn-rom                                                           FAIL
vfinal                                                                   ERROR
[####################################################################] (19/19)

************* SUMMARY *************
Passed   5/19 tests
Failed   2/19 tests
Errors  12/19 tests
```

I note that the `ccomp` tests are being removed in #499, but maybe we should re-evaluate that decision?

Finally, I figured out a bit more about setting up vscode debugging, and it works for me (it triggers a breakpoint), both with `--debug` and without. So are you still having problems, @olzama?

olzama:
I do not currently have any problems with debugging but I haven't really tried much (just some random breakpoints).

I am currently having some problems with my new wh-tests, so, might report back (I haven't yet investigated them; some tests seem to be running infinitely; may or may not related to the changes, although the previous version of rtest did not have this issue. Then again, that was before merging.)



> I note that the `ccomp` tests are being removed in #499, but maybe we should re-evaluate that decision?

Those are my own tests; I am fairly sure I added all I wanted to add back then, so, these ones must be the ones I decided were testing the wrong thing or something like that.

olzama:
I am having the following problem with the new version:

```
(pydelphin-1.4.0) (base) MurkinPro:matrix-git olzama$ ./rtest.py 
Traceback (most recent call last):
  File "./rtest.py", line 780, in <module>
    exit_status = main(args)
  File "./rtest.py", line 119, in main
    return args.function(args)
  File "./rtest.py", line 135, in run_tests
    force=args.force or args.skipped,
AttributeError: 'Namespace' object has no attribute 'force'
```

Is that a python version issue perhaps? Which version do I need to run this? I have 3.7.3

dantiston:
@olzama it looks like `force` is an argparse object’s attribute that @goodmami is adding manually in some situations, but maybe missed the default? Shouldn’t be a version problem.

Which reminds me, is there a plan for rtest unit tests?

goodmami:
Yes sorry I took out the line because I had several changes that I wanted to commit separately then forgot to commit that line. I’ll push it soon.

goodmami:
> Which reminds me, is there a plan for rtest unit tests?

rtest is only for r(egression)test(s). Unit tests are run with `pytest` and they are located in `tests/unit/`. There's some issues with test discovery so I run it as follows:

```console
$ python -m pytest tests/unit/
```

Currently the choices tests all pass, the morphotactics ones all fail, and the toolbox-import one consumes all your memory and dies. Could be improved, I think.

dantiston:
@goodmami what I mean is unit tests of rtest.py itself to avoid issues like this one :-)

goodmami:
@dantiston Ah, oops. But I guess that's a good point: we shouldn't make the testing system so complex that it needs to be tested itself. Otherwise all our efforts will go toward engineering and maintaining the system to test the actual thing we care about.

Also (to all following) I pushed the changes so only explicitly skipped tests are skipped and incomplete tests are now errors. Running or listing the tests with `--skipped` now only runs/lists those explicitly skipped.

goodmami:
This generated much discussion and thus got bigger and more complex than I'd hoped, but in the end it's not a huge number of lines changed. All my current TODOs are done so shall we merge this now?

olzama:
Fine with me; thank you so much, Michael!

olzama (approved):

goodmami:
Ok, thanks, Olga.

@dantiston, if you really think we should write tests for our tests then please open a separate issue :) Aside from that I'll assume you're good with the current state of these changes.

dantiston:
@goodmami yes, these changes are great.

On testing, I think the current situation of trying to recover our regression tests and rebuilding a new system is evidence that regression tests need unit tests. Also, this particular case where the system was complex enough to have several cases that were tested but the main one wasn’t tested is also evidence that a system is large enough to require tests.

Anyway, yes, I will make a new issue. Unfortunately not much of anything has unit tests right now, but regression tests seem like a good first place.

olzama:
> I think the current situation of trying to recover our regression tests and rebuilding a new system is evidence that regression tests need unit tests

Maybe, but I am not so sure, at least at the moment.

I think it is more just an illustration of the fact that tests are actually quite expensive to maintain---which is why some programmers do not write any tests at all, ever, in principle.  There is a cost to that, too, obviously, and, in case with the Grammar Matrix, I think that the regression tests are absolutely crucial, for example. But tests for rtest? Might cost us more at the end. Also the Grammar Matrix unit tests: there is a reason why there are almost none; nobody has time to develop them, and if somebody did suddenly, I am really not sure the project has the capacity to maintain them... 

As it is, we spent a few days discussing the new changes (and there may be issues we haven't yet uncovered which we will find later, which is fine). I don't think that, had we had additional tests, this time would've been reduced dramatically. After all, we mostly discussed the nature of the changes, conceptually. The bug which happened at the very end took a second to fix, I bet (and yes, caused me to delay my work for a day or so, but that is usually fine, I have plenty of other stuff to work on, not to mention that I could have figured it out myself if I really wanted to).

olzama:
> we shouldn't make the testing system so complex that it needs to be tested itself

Amen! ;)

goodmami:
As mentioned, I'm skeptical the maintenance burden of testing our tests is worth it, and I don't think a unit test would've actually caught this particular issue. I'm still hoping to eventually have a GitHub action that runs the regression tests when commits are added to a pull request, and this might actually catch these sorts of problems because it runs the code in the repo and not the code on my hard drive.

dantiston:
I’m very surprised to hear the pushback against this extremely common industry standard. As a person who writes unit tests just about every day, I want to encourage y’all to reconsider your position, if you’re not writing unit tests on a regular basis. The grammar matrix currently basically doesn’t have unit tests, I believe, because there isn’t a culture of writing unit tests. There are basically no examples of unit tests for libraries which perpetuates the lack of unit tests for libraries. The system is very brittle (the live website hasn’t worked for me in several weeks) and developed almost exclusively by people with little engineering experience (read: 2014 T.J.). Unit tests will help make it less brittle.

In a standard testing tower, regression tests are a key pinnacle on top of a breadth of integration tests and unit tests. Unfortunately, with only relying on regression tests, many of the features of the matrix remain untested (I‘ve found several choices recently in the libraries that aren’t in the regression test choices). And when tests do fail, it is often very difficult to determine the cause. I’ve been working on converting the choices code over the last couple weeks by relying exclusively on the regression tests (the only passing unit tests aren’t applicable), it’s basically been an investigative procedure of trying to find what particular code is failing, sometimes several steps removed. Unit tests for these modules would short circuit these investigations.

And, just like regression tests, the way to fix the maintenance problem is to run them automatically after committing to a PR. Running the regression tests won’t check the add, remove, and update functionality of rtest.

Anyway, I’m fine to champion writing the unit tests. I’m developing the new choices to be 100% test coverage, including adding coverage to the uprev and informational code. I’ll look into setting up the unit tests to run on PRs (and make an issue).

olzama:
I believe this is a known schism (tests vs. no tests) :).  Which is to say, it is probably unlikely that either side will reconsider its position :). 

The only thing I'd like to reiterate is that I do not think there is a capacity to maintain unit tests in this project. There is barely a capacity to maintain this many regression tests either, so, I wonder what happens in the future, actually (when an RA coming in for a quarter spends 100% of their time on the regression tests). But at least the regression tests have linguistic meaning in them, so ultimately that is relevant to the task (even if the task ends up multiplied by something like 0.1). 

So, it is not that I disagree that tests work well for some projects, or even that they are useful in a general sense. It is just that I don't think it is realistic for us to have them here. 

> the way to fix the maintenance problem is to run them automatically after committing to a PR

No, that is not the maintenance issue, that is the actual use of the test as such. Maintenance is keeping the tests in sync with the advancing project and fixing bugs in the tests.

goodmami:
I've created #500 to better track the conversation lest this one (on a merged and closed PR) gets lost.

*edit*: and I will lock the conversation only as a reminder to please continue the conversation in #500.

PR #499, opened 2020-08-05
Remove skipped tests

olzama:
This is to remove the tests which are currently skipped, if we conclude they were meant to be removed (or if we conclude we can no longer recover what their purpose is).

PR #502, opened 2020-08-27
Merge wh-questions library in its current state

### LINKED ISSUE #506: div_particles and globals.py
There were some objections to introducing globals.py; also, there may be a way of computing the value for div_particles that is more consistent with what is done elsewhere in the code (e.g. a function in choices.py which gets called appropriately).

olzama:
OK....

I would like to merge the wh-library in now. This PR is huge, and I am not sure anybody will want to review it (which is fine). 

Now that we started using Github, I suggested to Emily that the new libraries are no longer developed in the previous fashion, when the person would work on their library for months if not years. That results in overwhelmingly difficult merges and makes careful review virtually impossible.

Perhaps in the future, people work on short self-contained branches, one (sub)phenomenon at a time, whichever the minimal work item could be defined without breaking tests, and then comfortably merges into the trunk, not worrying about the website and the actual release of the library in full. The trunk can be separate from the website.

Now, this PR contains lots of various changes, most having to do with wh-questions but some having to do with things like ICONS updates and making verbs SPR-underspecified (and lots of other things SPEC-empty). 

The regression tests have been updated mostly with respect to ICONS, as I fixed some of formerly broken appends and the ICONS started showing up. Conversely, I fixed a type in matrix.tdl which contributed ICONS which did not yet make sense, so, I got rid of those underspecified ICONS in some 50 tests.

There is (at least) one regression which I will not fix now (because I do not know how) but will file an issue: it is item 197 in the Russian wh test (wh-rus-dev). There is a new ICONS there which I don't want, but similar things have always been in other wh-tests. Not sure why this one is new. Must be a formerly broken append as well. Now, I don't want any of them, but they will stay for now.

There is still plenty of work to do. For example, I need to simplify the WH feature to only contain an OR (it currently also has an unused AND). I will file an issue.

The information structure in wh-questions currently is not right. Again, will file an issue and hopefully will address it some time soon.

But I want to merge this in and start working on short branches instead.


goodmami:
Yes, sorry, I don't have time to review this. Even if you ignore the 1218 regression test files and only consider the 25 or so Python/TDL files that changed.

I agree that smaller PRs are a good goal. Multi-month or year efforts by a student might be better tracked by a project board instead of a giant branch. Also we might need to be explicit about our branching model, whether it's [GitFlow](https://nvie.com/posts/a-successful-git-branching-model/), [GitHub Flow](https://guides.github.com/introduction/flow/), or something else. E.g., do we need separate "trunk" and "live" branches? A changelog could help us understand what really has changed, especially when the Git commit messages aren't informative or easily skimmable. We probably should open an issue to discuss all these topics, however.

dantiston:
I could review it this weekend.

I agree with the branching strategy concerns and opening an issue to discuss. I do worry that a CLMS student with little to no software engineering experience (read 2014 T.J.) will struggle with it, but to our earlier conversation, following industry best practices will be good experience.

olzama:
Thanks, @dantiston, I appreciate it!

With this PR, I think it will be more effective to merge it as soon as possible and start working on small branches to address any issues that I already know of or that might be identified in review. But I can imagine there being some kind of changes that it would make sense to do first (perhaps something along the lines of missing parses in regression tests, or something else completely unexpected). Any analysis issues or style issues will have to be merged in, filed as issues, and addressed later, I think. Meaningless ICONS also (I got rid of some but there is more to do).

goodmami:
For the best-practice discussions, see #503 and #504.

gmcs/choices.py
```diff
@@ -2292,6 +2328,31 @@ def convert_31_to_32(self):
             if strat.get('possessum-type')=='non-affix':
                 if not strat.get('possessum-mark-order'):
                     strat['possessum-mark-order'] = hc
+
+    def convert_32_to_33(self):
+        pass
```
dantiston: I think this `pass` doesn't do anything? It's confusing.

gmcs/choices.py
```diff
@@ -2292,6 +2328,31 @@ def convert_31_to_32(self):
...
+        if self.get('q-part') == 'on':
+            orth = self.get('q-part-orth')
+            self.delete('q-part-orth')
+            self['q-particle'] = ChoiceList(full_key='q-particle')
+            cdict = ChoiceDict(full_key = 'q-particle1')
```
dantiston, line 2350: These `full_key`s look wrong, but I'm definitely not an expert on the current choices files (and I really need to finish getting rid of them 🙂).

gmcs/constants.py
```diff
@@ -33,3 +33,26 @@
 OV_ORDERS = ['sov', 'ovs', 'osv', 'v-final']
 VFINAL = ['sov','osv','v-final']
 VO_ORDERS = ['svo', 'vos', 'vso', 'v-initial']
+
+### WH-QUESTIONS
+
+IN_SITU = 'in-situ' # No question phrase fronting
+EMB_INSITU = 'embed-insitu' # Embedded in situ questions (when fronting is also possible; that's rare)
```
dantiston: I don't think we have a standard for constants names, but I feel like some of these are hard to understand, and could be improved by being replaced with the full words, except like `WH` (which is a library name) and `OBLIG`, `MULTI`, and `INFL` which are pretty standard. This one in particular seems problematic for namespace collisions with Emily's initials 🙂 

On a more meta level, I remember you suggesting using constants, and while in general I agree, I think constant naming is a problematic issue. Constants add a (often useful) layer of abstraction that can also be a layer of confusion with confusing or misleading names. Name spacing is also hard, and while I don't think it needs to be solved perfectly in this PR, it would be nice to have a strategy, for instance defining library constants in library files (my preferred), using prefixes for library-specific constants, etc.

gmcs/customize.py
```diff
@@ -522,9 +528,12 @@ def customize_matrix(path, arch_type, destination=None, force_dest=False):
     valence_change.customize_valence_change(mylang, ch, lexicon, rules, lrules, hierarchies)
     word_order.customize_word_order(mylang, ch, rules)
     coordination.customize_coordination(mylang, ch, lexicon, rules, irules)
-    yes_no_questions.customize_yesno_questions(mylang, ch, rules, lrules, hierarchies)
     clausalmods.customize_clausalmods(mylang, ch, lexicon, rules, roots, trigger)
     clausalcomps.customize_clausalcomps(mylang,ch,lexicon,rules)
+    adverbs_adpositions.customize_adv_adp(ch,mylang,rules)
+    wh_ques.customize_wh_ques(mylang,ch,rules,roots)
+
+    # Service customization
```
dantiston: What does this comment mean?

gmcs/feature_type_use.py
```diff
@@ -0,0 +1,7 @@
+# Attempt to flag specific features which were added to the grammar by a component.
+# One such feature is INIT; it is not needed in all grammars but can be added by the Word Order library,
+# the clausal complements library, the clausal modifiers library, and the Wh-questions library
+# (and possibly other libraries).
+
+USED_FEATURES = {'INIT':False}
+USED_TYPES = {'qdet':False,'qpro':False,'qadv':False,'qverb':False}
```
dantiston, line 7: I like this idea in principle, but I don't think this will work in practice like this forever? The way regression testing happens right now with `multiprocess`/`subprocess`, I guess it should be fine, but if someone tried to customize two grammars with one Python instance, it wouldn't reset between customizations.

I think there's a lot of opportunity for refactoring the current system, and this sort of thing seems to fit into that, so I'll open an issue for this.

gmcs/globals.py
```diff
@@ -0,0 +1,3 @@
+### Global variables
+# Diverse question particles: some can occur with polar but not with wh-questions.
+div_particles = False
```
dantiston: Ditto comment on the `feature_type_use` file, but global state in general is often considered bad practice (e.g. https://softwareengineering.stackexchange.com/questions/148108/why-is-global-state-so-evil), and I'd rather see us moving away from it instead of moving towards it.

gmcs/linglib/adverbs_adpositions.py
```diff
@@ -0,0 +1,31 @@
...
+   NON-HEAD-DTR.SYNSEM.LOCAL.CAT [ HEAD +jrp, VAL.COMPS < > ] ].
+ '''
+
+ADJ_HEAD = '''my-adj-head-phrase := adj-head-int-phrase &
+ [ HEAD-DTR.SYNSEM [ LOCAL.CAT [ HEAD +nvr ] ],
+   NON-HEAD-DTR.SYNSEM.LOCAL.CAT [ HEAD +jrp, VAL.COMPS < > ] ].
+ '''
+
+def customize_adv_adp(ch, mylang, rules):
+    if len(ch.get('adv',[])) > 0 or len (ch.get('normadp',[])) > 0: #need to handle also adjectives here
```
dantiston: Drop the space after `len`, also, I agree about handling adjectives. Aren't these rules in `matrix.tdl` already? I guess these are just adverb and adposition subtypes in `mylang`?

gmcs/linglib/clausalcomps.py
```diff
@@ -613,6 +628,28 @@ def customize_clausal_verb(clausalverb,mylang,ch,cs,extra):
...
 
+    if not is_nominalized_complement(cs):
+        if cs['ques'] == 'prop':
+            mylang.add(clausalverb + ' := [ SYNSEM [ LOCAL.CAT.VAL.COMPS < [ LOCAL [ CAT.WH.LOGICAL-OR.BOOL -, '
+                                     '                                               CONT.HOOK.INDEX.SF prop ] ] >,'
+                                     'NON-LOCAL.QUE 0-alist ] ].'
+                       , merge=True)
+        elif cs['ques'] == 'ques':
+            from gmcs.constants import MTRX_FRONT
+            from gmcs import globals
```
dantiston: Unused import?

gmcs/linglib/clausalcomps.py
```diff
@@ -613,6 +628,28 @@ def customize_clausal_verb(clausalverb,mylang,ch,cs,extra):
...
+            mylang.add(clausalverb + ' := [ SYNSEM.LOCAL.CAT.VAL.COMPS < [ LOCAL.CONT.HOOK.INDEX.SF ques ] > ].'
+                       , merge=True)
+            from gmcs.constants import SINGLE,MULTI
```
dantiston: 1. this should at the very least be combined with the import on 638 (4 lines above here)
1. why not put these are the top of the file?

gmcs/linglib/features.py
```diff
@@ -74,6 +82,9 @@ def customize_feature_values(mylang, ch, hierarchies, ch_dict, type_name, pos, f
         'xarg': 'LOCAL.CONT.HOOK.XARG.',  # XARG for adjectives
         'mod': 'LOCAL.CAT.HEAD.MOD.FIRST.',  # MOD for adjectives
         'comp': 'LOCAL.CAT.VAL.COMPS.FIRST.', # COMP for copulas
+        #'noun': 'LOCAL.CAT.VAL.SPEC.FIRST.', # olzama 2020-04-08 I think this was a bug.
+        # There was interference between this line for determiner inflection and prefix_map
+        # for full-form determiners.
```
dantiston, line 87: The diff is showing you adding all of this? Should it just be deleted or maybe change the message?

gmcs/linglib/lexical_items.py
```diff
@@ -40,6 +47,9 @@ def adp_id(item):
     """Return the identifier for an adposition lexical item."""
     return get_name(item) + '-adp-lex'
 
+def qpart_id(item):
+    """Return the identifier for a determiner lexical item."""
```
dantiston: Bad comment

gmcs/linglib/lexical_items.py
```diff
@@ -314,7 +326,49 @@ def customize_verbs(mylang, ch, lexicon, hierarchies):
...
+            create_interrogative_verb_type(cases,ch,hierarchies,lexicon,mylang,verb)
+
+def create_interrogative_verb_type(cases,ch,hierarchies,lexicon,mylang,verb):
+    vtype = verb_id(verb)
+    # clausal verb's valence and its complement's head constraint:
+    vtype,head = clausalcomps.update_verb_lextype(ch,verb,vtype)
+    if verb['predtype'] in ['manner','loc']:
+        mylang.add(vtype + lexbase.ITRG_FOUR_REL)
+    elif verb['predtype'] == 'ref':
+            mylang.add(vtype + lexbase.ITRG_THREE_REL)
```
dantiston: Bad indentation

gmcs/linglib/lexical_items.py
```diff
@@ -340,20 +394,20 @@ def create_verb_lex_type(cases, ch, hierarchies, lexicon, mylang, verb):
...
+        TDLencode(name) + ' := ' + stype + ' & \
                     [ STEM < "' + orthstr + '" >, \
                       SYNSEM.LKEYS.KEYREL.PRED "' + pred + '" ].'
     lexicon.add(typedef)
 
 
 def construct_supertype_names(cases, ch, stype_names, verb):
     val = verb.get('valence')
-    if val:
+    if not val == '':
```
dantiston: Why this change? Are you now expecting `val == None` to work?

gmcs/linglib/lexical_items.py
```diff
@@ -970,7 +1087,6 @@ def customize_cops(mylang, ch, lexicon, hierarchies, trigger):
                                      COMPS < [ LOCAL.CAT cat-sat & [ HEAD.PRD +,
                                                            VAL [ SUBJ < >,
                                                                  COMPS < > ] ] ] >,
-                                     SPR < >,
```
dantiston, line 973: Was this to support something with adverbs? Why wouldn't one want a copula to be `SPR <>`?

gmcs/linglib/lexical_items.py
```diff
@@ -1014,6 +1130,30 @@ def customize_cops(mylang, ch, lexicon, hierarchies, trigger):
...
+    if ch.get('has-aux') == 'yes':
+        mylang.add('norm-adposition-lex := [ SYNSEM.LOCAL.CAT.HEAD.MOD < [ LOCAL.CAT.HEAD.AUX - ] > ].')
+    supertype = 'norm-adposition-lex'
+    for adp in ch.get('normadp'):
+        typename = adp.full_key + '-' + supertype
+        mylang.add(typename + ' := ' + supertype + '.')
+        for stem in adp['stem']:
+            add_stem_to_lexicon(lexicon,stem,typename)
+        features.customize_feature_values(mylang,ch,hierarchies,adp,typename,'normadp')
+        if not feature_type_use.USED_FEATURES['INIT']:
```
dantiston, line 1147: I'm not sure I understand the point of these `feature_type_use` checks. Is it to avoid avoid adding the same thing to `mylang.tdl` twice? Doesn't the TDL library already handle duplicates?

gmcs/linglib/yes_no_questions.py
```diff
@@ -107,49 +136,116 @@ def customize_yesno_questions(mylang, ch, rules, lrules, hierarchies):
...
+            mylang.add('non-ynq-word := word-or-lexrule & [ SYNSEM.NON-LOCAL.YNQ.LIST < > ].')
+            mylang.add('basic-zero-arg :+ non-ynq-word.',section='addenda')
+            mylang.add('basic-one-arg :+ non-ynq-word.',section='addenda')
+            mylang.add('basic-two-arg :+ non-ynq-word.',section='addenda')
+            mylang.add('basic-three-arg :+ non-ynq-word.',section='addenda')
+            mylang.add('intersective-mod-lex :+ non-ynq-word.',section='addenda')
+            roots.add('root := [ SYNSEM.NON-LOCAL.YNQ.LIST < > ].')
+            mylang.add(INT_CL,section='phrases')
+            mylang.add(DECL_CL,section='phrases')
+            rules.add('int-cl := int-cl.')
```
dantiston, line 212: This looks bad?

gmcs/linglib/wh_ques.py
```diff
@@ -0,0 +1,312 @@
...
+from gmcs.utils import get_name,TDLencode, orth_encode
+
+from gmcs import globals
+
+
+'''
+CONSTANTS
+'''
+
+WH_Q_PHR_NO_OR_SG_OBLIG_MULTI = ''' wh-ques-phrase :=
```
dantiston, line 13: Several of these definitions have spaces at the beginning, but many do not. It seems no space would be preferable.

gmcs/linglib/wh_ques.py
```diff
@@ -0,0 +1,312 @@
...
+        if ch.get(MTRX_FR_OPT) == ALL_OBLIG:
+            mylang.add(WH_Q_PHR_NO_OR_SG_OBLIG_MULTI) # QUE.LIST is empty
+        # Rule out structural ambiguity for sentences like "Who sleeps where?"
+        if ch.get('word-order') in ['svo', 'sov', 'osv' ]:
+            mylang.add('''head-adj-int-phrase :+ [ HEAD-DTR.SYNSEM [ L-QUE -,
+             LOCAL.CAT.VAL [ SUBJ clist, COMPS clist ] ] ].''',section='addenda')
+        if ch.get('word-order') == 'free':
+            mylang.add('''adj-head-int-phrase :+ [ HEAD-DTR.SYNSEM.LOCAL.CAT.VAL.SUBJ clist ].''', merge=True)
+            mylang.add('''head-adj-int-phrase :+ [ HEAD-DTR.SYNSEM.LOCAL.CAT.VAL.SUBJ clist ].''', merge=True)
+        # The below does not make sense?
```
dantiston: ?

matrix-core/matrix.tdl
```diff
@@ -228,12 +232,14 @@ non-local-min := avm.
 non-local := non-local-min &
   [ SLASH append-list,
     QUE append-list,
-    REL append-list ].
+    REL append-list,
+    YNQ append-list ].
```
dantiston, line 241: No comment for YNQ?

matrix-core/matrix.tdl
```diff
@@ -958,9 +986,111 @@ ocons := olist & cons &
...
+  [ OTHER-BOOL #result,
+    RESULT-BOOL #result ].
+
+my-and := *top* &
+  [ TEST1 [ AND < [ BOOL + ], [ BOOL + ], [ BOOL + ], [ BOOL + ] > ],
+    TEST2 [ AND < [ BOOL + ], [ BOOL - ] > ],
+    TEST3 [ OR < [ BOOL + ], [ BOOL - ] > ],
+    TEST4 [ OR < [ BOOL - ], [ BOOL - ] > ],
+    TEST5 [ NOT [ BOOL - ] ],
+    TEST6 [ OR < [ BOOL - ], [ NOT [ BOOL - ] ] > ] ].
```
dantiston: Delete?

tests/regression/regression-test-index
```diff
@@ -477,5 +477,35 @@ ccomp-yak=Clausal complements held-out: Yakima Sahaptin (yak)
 adj-yes-no-cop-aux-inv=auxiliaries and copulas intersecting with yes-no library
 adj-yes-no-cop-inv=copulas intersecting with yes-no library
 subj-aux-inv-q=yes-no questions with subject auxiliary inversion
-adj-yes-no-cop-aux-inv=auxiliaries and copulas intersecting with yes-no library
-adj-yes-no-cop-inv=copulas intersecting with yes-no library
```
dantiston: These are working in trunk

tests/regression/skeletons/adnom-poss-spec-possessor-affix-two-strat-mod-spec-var/item
```diff
@@ -1,3 +0,0 @@
-1@unknown@formal@none@1@S@det noun1possessor noun2 itr-verb3sg@@@@@@@ekayen@2018-01-13
```
dantiston: Why are these being deleted?

tests/regression/home/gold/subj-v-inv-obj-drop/result
```diff
@@ -1,6 +1,4 @@
...
...94 head-comp 0.000000 0 6 (89 head-comp 0.000000 0 3 (86 inv-lr 0.000000 0 1 (85 3sg-suffix 0.000000 0 1 (7 do 0.000000 0 1 ("Dos")))) (88 head-spec 0.000000 1 3 (8 the 0.000000 1 2 ("the")) (87 singular-lex 0.000000 2 3 (9 dog 0.000000 2 3 ("dog"))))) (93 head-comp 0.000000 3 6 (90 inf-suffix 0.000000 3 4 (10 chase 0.000000 3 4 ("chase-inf"))) (92 head-spec 0.000000 4 6 (11 the 0.000000 4 5 ("the")) (91 singular-lex 0.000000 5 6 (12 cat 0.000000 5 6 ("cat")))))))@@("S" ("S" ("V" ("V-M" ("V" ("V" ("Dos")))) ("NP" ("D" ("the")) ("N" ("N" ("dog"))))) ("VP" ("V" ("V" ("chase-inf"))) ("NP" ("D" ("the")) ("N" ("N" ("cat")))))))@[ LTOP: h0 INDEX: e2 [ e SF: ques E.TENSE: tense E.ASPECT: aspect E.MOOD: mood ] RELS: < [ "_def_q_rel"<-1:-1> LBL: h4 ARG0: x3 [ x SPECI: bool COG-ST: cog-st PNG.PER: 3rd PNG.NUM: sg ] RSTR: h5 BODY: h6 ]  [ "_dog_n_rel"<-1:-1> LBL: h7 ARG0: x3 ]  [ "_chase_v_rel"<-1:-1> LBL: h1 ARG0: e2 ARG1: x3 ARG2: x8 [ x SPECI: bool COG-ST: cog-st PNG.PER: 3rd PNG.NUM: sg ] ]  [ "_def_q_rel"<-1:-1> LBL: h9 ARG0: x8 RSTR: h10 BODY: h11 ]  [ "_cat_n_rel"<-1:-1> LBL: h12 ARG0: x8 ] > HCONS: < h0 qeq h1 h5 qeq h7 h10 qeq h12 > ICONS: < > ]@((:ascore . 0.0) (:probability . 1.0))
```
dantiston, line 4: This seems odd. The `parse` file used to have 15 items and now has 15 items. But the result used to only have 6 and now only has 4?

dantiston (changes requested):
It looks like this includes #499? Is that intentional?

I think in general this looks great. I certainly agree that it is a very large PR and could be broken into many chunks. I think some of those chunks are not related to the wh-questions work (to my knowledge), and I do think there's some work to do on thinking and maybe some infrastructure for working on partial features, which could be great.

I reviewed all of the Python and the TDL. I glanced through the tests, but I didn't look at many that were wh-questions as I trust those were correct.

I tried to find the ICONS changes in the tests, and the ones that I found looked great.

My primary concern is the deletion of tests. Many of the deleted tests are clausal complement tests, so I'm assuming there's some knowledge there, but there's a bunch of deleted adnominal possession tests and some other ones (e.g. `neg1b-neg2fd`), too? Should these deleted test just be marked skipped if they're not working? But I think they are working on trunk?

My secondary concern is the new globals. There are already (true) globals in the code that could be removed, but adding new ones seems mediocre. I understand there may be a philosophical divide here, but I'm sort of confused about their purpose? It seems you're trying to avoid adding the same feature twice, but that shouldn't be a big deal? Also, other libraries already do this (going back to SFD days) by storing choices in the choices file. Something like this seems preferable. In the choices refactor work I've done, I've thought that it would be handy to have a copy of the user's input choices and an "enriched" choices, and maybe this would accomplish what you're going after? As I mentioned in the comment though, this can be addressed later.

Also, I'm noticing the way GitHub formats the inline comments is not what I was expecting. Many of the comments are on the final line of the preview, but many were for several lines. Hopefully it is discernible.

tests/regression/home/gold/subj-v-inv-obj-drop/result
dantiston: Ah, I see your comment in your commit. I'm still confused about why there are 16 parses and 4 MRS, though.

tests/regression/regression-test-index
olzama: Oh shoot, these keep getting deleted every time I do *anything* now. So confusing! Did not mean to delete them at all (unlike a bunch of causal complements and adnominal possession tests which were lacking things and would be too many to skip).

olzama: Ooh OK they did not get deleted after all. It was a double entry (for some reason). The tests are there (including in the index).

gmcs/customize.py
olzama: Don't know. I doubt I wrote it; more likely it got reshuffled up or down from somewhere...

olzama:
> It looks like this includes #499? Is that intentional?

I forgot that I had never merged in #499! I did that just now.

olzama:
> My primary concern is the deletion of tests. Many of the deleted tests are clausal complement tests, so I'm assuming there's some knowledge there, but there's a bunch of deleted adnominal possession tests and some other ones (e.g. `neg1b-neg2fd`), too? Should these deleted test just be marked skipped if they're not working? But I think they are working on trunk?

See #499 for these. If they used to pass in the trunk, that would be by mistake (empty gold or something like that). All these tests either are mostly empty or are my own tests which I know I wanted to delete. Emily and I agreed there is no reason to keep them around in the hopes that someone fixes them one day; too many SKIPPED tests is also bad and should be reserved for conscious decisions. If a test seems to be missing most of its content, there is probably little reason to keep it around.

olzama:
> My secondary concern is the new globals. There are already (true) globals in the code that could be removed, but adding new ones seems mediocre. I understand there may be a philosophical divide here, but I'm sort of confused about their purpose? It seems you're trying to avoid adding the same feature twice, but that shouldn't be a big deal? Also, other libraries already do this (going back to SFD days) by storing choices in the choices file. Something like this seems preferable. In the choices refactor work I've done, I've thought that it would be handy to have a copy of the user's input choices and an "enriched" choices, and maybe this would accomplish what you're going after? As I mentioned in the comment though, this can be addressed later.

Yeah, this is by all means unfinished (and yes, a philosophical divide may emerge in a discussion like this, I suppose!). 

What I was after is not having to **compute** a value for a feature more than once. There are decisions which need to be made based on a combination of choices, for example: are there question particles which exhibit diverse behavior with respect to what type of questions they occur with? Having to compute that twice (or more than once) is not ideal. Now, what I strongly believe needs to be avoided is duplicate code (because typos and bugs); but what could be done instead of the "globals" is a function that gets called in all places. Then the value would be computed more than once but at least by the same place in the code. Let's open an issue about this.

Yes, "enriched choices" sounds like something I am after here, perhaps.

gmcs/choices.py
olzama: Wow. I never noticed! Thanks :).

gmcs/choices.py
olzama: What do you mean by "look wrong"?

gmcs/constants.py
olzama: I think constants must be used for the sake of minimizing typos (of which we have lots and lots, and occasionally it costs hours of painful debugging).

As for what they should be -- that is of course another matter. The familiar tension between line length and explicitness...

gmcs/choices.py
olzama: I removed the rogue *pass* statement :).

gmcs/feature_type_use.py
olzama: I wanted to start collecting some statistics about what features are used how. I did not think about it much in terms of doing it over multiple grammars simultaneously although it would of course be interesting. But even within one grammar is interesting enough.

Of course I only started it and didn't really do anything interesting with these dicts yet.

gmcs/globals.py
olzama: The question of the meaning and consequences of something being declared a "bad practice" aside, I think in this case I am after something slightly different (which I also described above). 

This value should only be computed once and then never change. I will file an issue and will later look into why I was not able to do this in the same fashion as other things are dealt with by choices.py (has_adp_case() ) etc. There may or may not have been a good reason for it. But I want to close this PR first, as it is just too big to think about.

I think what I will do right now is rename this from "global variables" to "complex choices" or something like that.

olzama: As a side note, having read the linked discussion, I note that all of our variables are "global" in the sense that the discussion is... discussing. Meaning anyone can change the value of ch[x] at any point, and python passes things by reference. So what they are discussing there doesn't truly apply to our paradigm and data structures; we are not in a situation where we are protected from a change of state in the first place.

olzama: I went ahead and changed this to just call a function from choices.py (consistent with what is done in other places). Just doing it once would perhaps be better but that's a topic for another time ("enriched choices").

gmcs/linglib/adverbs_adpositions.py
olzama: The matrix.tdl head-adjunct type is too general and overgenerates, were I to use it for my tests. So now there is a work item there, to integrate this with what other libraries do for head-adjunct (information structure, adjectives, some others). All the tests now pass but that's an illusion of harmony; in reality, were I to add adjectives to my wh tests, things would fall apart.

But the bottom line is: other libraries add a simple head-adjunct rule with no further constraints, and that doesn't really work (overgenerates).

gmcs/linglib/clausalcomps.py
olzama: It was used just below. Anyway, I moved the import statement to the top.

gmcs/linglib/features.py
olzama: I commented it out but was hesitant to delete... I was not sure I was doing the right thing there (and still am not). Not sure how to change the message to make it more informative...

gmcs/linglib/lexical_items.py
olzama: So... This was a long time ago, and I don't remember really, but this was actually crashing on me in some cases until I made this change. But, I wouldn't be able to reproduce it, I don't think...

olzama: I am going to assume that this was something weird perhaps having to do with python2... I am going to just change this back, assuming there will be no change in the tests.

gmcs/linglib/lexical_items.py
olzama: No verbs should be SPR < > (they all were! spuriously) because adpositions should be able to select for nonemtpy SPR things (otherwise they attach below and above NPs).

gmcs/linglib/lexical_items.py
olzama: You are right that this is not really serving any purpose yet. This is to start and collect statistics about which features are used by which libraries---for which this dict needs to be extended. I suppose I could clear this out from the trunk and start a develop branch for this instead.

olzama: ...Or not. Perhaps this is an example of a minimally working thing that can be merged into the trunk even if it is not serving any purpose (as opposed to it sitting in some branch for months).

gmcs/linglib/yes_no_questions.py
olzama: What does?

matrix-core/matrix.tdl
olzama: I don't know that one is needed. There are too many comments in matrix.tdl as it is...

matrix-core/matrix.tdl
olzama: These are examples of usage that Guy created; I think they are helpful, so I'd keep them.

tests/regression/skeletons/adnom-poss-spec-possessor-affix-two-strat-mod-spec-var/item
olzama: #499

tests/regression/home/gold/subj-v-inv-obj-drop/result
olzama: Don't know!..

olzama:
Alright, many thanks, @dantiston for the review. I addressed some things he pointed out and opened a couple issues for the remaining things which I do not want to address now. I will now finally merge this PR in and hopefully we will never see a PR of this size here ever again!! :)

gmcs/choices.py
goodmami: In #472 (still a draft) I removed the `full_keys()` method but not the `full_key` initialization parameter, but I expect it, too, will go away in the end.

Style-wise the spaces around the `=` are not PEP-8 compliant (but so far I haven't been raising style issues because there are so many in the codebase already... these could be fixed in a style-fixes-only kind of PR).

gmcs/constants.py
goodmami: The comment helps, but I agree that constants (and variable names in general) should generally be more descriptive, and `EMB` frequently occurs in comments as Emily's initials so I was confused here, too. `EMBED_IN_SITU` is only 3 more characters.

gmcs/customize.py
goodmami: Looks like it was introduced in c7068227bdcd9a20d608ba43e9fa13c5b7efbf6f and wasn't elsewhere before. Maybe you just forgot why you wrote it? In which case it's probably doing no good and can be removed.

gmcs/feature_type_use.py
goodmami: I think T.J. means that in a multithreaded (not multiprocess) use of the Matrix (which we do not currently do but in the future who knows) this global might be modified by different customizations in the same Python process, leading to incorrect results. But, as a side-point, if these are intended to be modified they should not be ALL_CAPS as that indicates they are constants.

gmcs/linglib/features.py
goodmami: The past-tense ("this was a bug", "There was") suggests this code was previously in use. Since it's new and unused code, it should be framed as "Were someone to try and do X, they might use something like this: ..., but in initial testing it failed because Y so it is not currently used", or something. But I'd prefer removing it entirely to avoid the confusion.

gmcs/linglib/lexical_items.py
goodmami: > thing that can be merged into the trunk even if it is not serving any purpose

I'd think we'd want things not serving any purpose excised from trunk.

gmcs/linglib/yes_no_questions.py
goodmami: `int-cl` is its own supertype? How do these grammars even compile with ACE? Maybe the line is not used in our tests?

matrix-core/matrix.tdl
goodmami: They will be added to every Matrix grammar customized and they are not meant to be used. I think they belong in documentation.

tests/regression/regression-test-index
goodmami: Using `rtest.py add` should recreate the index, so if we have (manually inserted) duplicates they will be removed.

goodmami:
Thanks @dantiston for doing the review. I added some comments to issues you've raised. I agree that we should not introduce new globals and instead should work towards removing any that exist.

gmcs/linglib/features.py
olzama: That's exactly what this is: I commented this out because it was causing me problems, but I am not sure yet I should just delete it. It is not newly added code.

gmcs/linglib/lexical_items.py
olzama: How is that consistent with the idea of incremental merges of unfinished libraries though?

gmcs/linglib/yes_no_questions.py
olzama: That's rules though; I don't think there is a restriction of this sort there? That's just a name for the rule to be used? I mean I can rename it but I don't think it is causing any problems.

matrix-core/matrix.tdl
olzama: Documentation of what though?

gmcs/feature_type_use.py
olzama: (another reason to not use multithreading, as far as I am concerned :) ).

matrix-core/matrix.tdl
goodmami: Maybe here: http://moin.delph-in.net/MatrixTop

Or something similar to this: http://moin.delph-in.net/GeFaqDiffList

Or at least comment out the TDL with a note saying that it's an example and what it's exemplifying. It's not really clear what purpose this serves as it is.

gmcs/feature_type_use.py
goodmami: We're not using multithreading

gmcs/linglib/features.py
goodmami: It appears newly-added compared to trunk, although it may have been in your branch for a long time. It's not currently in trunk.

gmcs/linglib/lexical_items.py
goodmami: Are you referring to my proposal in #503? If so then I think I understand your comments above better. But to clarify, merging partial things into trunk would be better if they (1) were in their own branch and not as a rider on some other set of changes, (2) had some issue associated with them (or at least a clear and documented goal of what the changes are working toward), and (3) don't break anything. I haven't tested this, but I think it satisfies (3) but not (1) or (2). I'm concerned about experimental features being added and forgotten about. If they are grouped with some set of meaningful changes, it is hard for someone later to tease apart what is what.

That said, I like the idea of being able to quantify which features are used by which libraries, but I think it should have an issue and a branch since that effort seems orthogonal to the point of the current PR.

gmcs/linglib/yes_no_questions.py
goodmami: Ah ok, so it's not creating a type. I was unaware that rules and types have separate namespaces (or do they?). Perhaps we should adopt some more thorough naming conventions for rules and types like the ERG has, because even if this works it is still a bit confusing.

matrix-core/matrix.tdl
olzama: OK sounds good! Will do the last option for now, I think.

gmcs/linglib/yes_no_questions.py
olzama: I can rename it if that's confusing. Didn't occur to me.

gmcs/feature_type_use.py
olzama: I know :). Good :). (Sorry, I am just messing with ya at this point)

gmcs/customize.py
olzama: OK, sounds good!

gmcs/linglib/features.py
olzama: Weird... I am 99% sure it's somebody else's code. But I can remove it of course. Just not entirely sure about this particular one (and what its purpose was originally)

matrix-core/matrix.tdl
dantiston: I agree there are a lot of spurious comments in `matrix.tdl`, but basic features seem maybe the best thing to document. Also, while I don't know if I would say `SLASH` and `QUE` are the most transparent names, I can get there. I have no idea what `YNQ` stands for.

But, maybe that doesn't matter. Up to you.

gmcs/choices.py
dantiston: @olzama Oh, now that I look at it again, I think I misunderstood what these were doing. These are top level values, so I guess they're correct.

@goodmami in the Choices redo (still not published...) that I've been working on, I have kept `full_key`. I don't think it's inherently bad (though sometimes used poorly), and it is used *a lot* in the code.

gmcs/linglib/lexical_items.py
olzama: I was referring to what I suggested to Emily, that it is ultimately detrimental to have any work go unmerged for a long time. And that it would be best to merge more or less anything so long as it doesn't break the tests and makes some sort of sense---and have releases in addition to that (and be more scrupulous about those). I don't think lengthy discussions of whether something is fit for merge are ultimately helpful but some limited discussion certainly is! In this case, I am not sure whether it is more likely for this to get lost in the trunk or in a branch that isn't actively being developed (and to my knowledge, there is nobody in particular that is going to pick up this work soon...). That said, I will certainly open an issue for this one.

goodmami:
@olzama you've linked #506 to this pull request, meaning that it will be closed when this is merged. Is that what you want?

gmcs/constants.py
dantiston: @olzama definitely agree constants are great. I just think these names could be a little longer (as @goodmami suggests).

gmcs/choices.py
goodmami: @dantiston Ok I did `git grep full_key` and yes, it's more prevalent than I thought. Consider merging #472 to get rid of `full_keys()` method, which was only (ab)used in the adnominal_possession library. I still think most uses of `full_key` are not great. I was hoping we'd just use TOML and use the decoded dict instead of a custom object, but if we're using `full_key` it might not be that simple.

gmcs/linglib/lexical_items.py
olzama: Oh, and this PR definitely contains lots of stuff! Which is not great but it is what it is. No way of going back and turning this into a bunch of separate branches now :).

dantiston: fwiw, my personal preference is I think just a little beyond what @olzama is suggesting. I'm happy to see something merged if it doesn't break anything *and* it does something useful. I recognize "doing something useful" is a vague concept, but I think saying this particular feature's useful thing is calculating stats is not currently accurate, so I wouldn't say it passes that bar. But there's definitely a trade-off...

matrix-core/matrix.tdl
olzama: Yes/No Question :) Much more transparent that SLASH, in my opinion :) 
I can add one comment regarding what YNQ is, but I think I won't add it everywhere where it occurs.

olzama:
> @olzama you've linked #506 to this pull request, meaning that it will be closed when this is merged. Is that what you want?

Yes; I addressed that already!

gmcs/linglib/yes_no_questions.py
dantiston: So `int-cl` isn't its own supertype? The LHS is a definition in `rules.tdl` and the RHS is a type specified in `mylang.tdl`? I agree this is confusing, and renaming would be helpful if possible.

gmcs/linglib/wh_ques.py
dantiston: Oh, I'm sorry, I wasn't specific enough. I don't know what the right answer is for indentation of the feature structure. What I was referring to was for instance here on line 13 you have the following with a space at the beginning of the string:
```
WH_Q_PHR_NO_OR_SG_OBLIG_MULTI = ''' wh-ques-phrase :=
                                   ^
```

But on line 27 you have, with no space:
```
MAIN_WHQ = '''main-wh-ques-phrase := wh-ques-phrase &
```

gmcs/linglib/lexical_items.py
goodmami: > it would be best to merge more or less anything so long as it doesn't break the tests and makes some sort of sense

We should consider the [technical debt](https://en.wikipedia.org/wiki/Technical_debt). Merging partial work in has a debt to finish it. The debt is higher if someone else (not the original author) needs to finish code, especially if they are unfamiliar with the code or even its purpose. Even if it's not finished, there's a debt to decide that the code is unneeded and to remove it. In other words, if it's not immediately helping the project and it's creating work for someone else, we should probably not merge it. Note that my opinions here are more about the sentiment expressed in the quote above than this particular bit of code because I'm not entirely sure what it's for and what state it's in.

> I am not sure whether it is more likely for this to get lost in the trunk or in a branch that isn't actively being developed

I guess the point is if it is going to get lost and forgotten, it's better to be lost and forgotten in a stale branch than in trunk. Having an associated issue would help it not be forgotten.

goodmami:
> Yes; I addressed that already!

Oh, sorry, I thought it was a new issue created as a response to this PR review.

dantiston (approved):
Thanks @olzama for making those changes! I think it looks good!

I think there are still some outstanding issues, such as renaming constants, changing the example type to a comment, some other pending clean-ups. But, it seems like there is consensus about what needs to change and how to change it, so I'll go ahead and accept! Please finish the clean-ups before merging.

olzama:
> > Yes; I addressed that already!
> 
> Oh, sorry, I thought it was a new issue created as a response to this PR review.

It was actually; I created it thinking I won't address it just yet but then went ahead and did. Sorry for the confusion.

gmcs/linglib/lexical_items.py
olzama: > I guess the point is if it is going to get lost and forgotten, it's better to be lost and forgotten in a stale branch than in trunk

But it is the opposite that is the point :). The point is to make it less likely that it is forgotten.

We have always had lots of unfinished work in the Matrix, and I don't see why it would change (that's the nature of the project; btw there are other things that I think are the nature of the project that have to do with things like conventions and protocols, but that's another topic). Given that we will always have lots of unfinished work, if we somehow make sure it always ends up in branches (which I don't think we will, so, it would also be inconsistent in the end: you'd expect such stuff to never be in the trunk but it still will be), we may end up with hundreds of branches a few years from now. My library alone would contribute at least a dozen of them if not more. Another library developer comes in (hopefully!) and a dozen more is added.

Honestly, I think in this case appropriately commented stuff in the trunk is ultimately preferable. 

Also, I am often informed by places in the code that I stumble upon, e.g. all the unfinished stuff in information structure has been important for me to look at as I was working in wh-questions. It would not have occurred to me to search the branches. It is more about the Matrix as a living organism, if you will. Sometimes you need to get a sense of what it's like as a whole, with all the unfinished parts. Branches, I feel, are excellent for active development but not for this kind of sense.

gmcs/linglib/features.py
olzama: I am not going to remove it for now. I am sure this code was here for a long time and almost equally sure I am not the one who had added it. No idea why it appears differently in the diff. This stopped working when I added lexical rules for determiners (e.g. CASE). And I am far from sure my solution for that is good, so, I'd rather not get rid of this trace of the previous situation. 

I understand the aversion to commented out code and I do not disagree in principle that it is best to not leave it in but in this case I will.

gmcs/customize.py
olzama: I changed the comment; I think a comment is in fact needed there, it is just that perhaps "Service" is not good.

gmcs/constants.py
olzama: I renamed the EMB-one but note that we have lots of issues in the code with long lines and inconsistent indentation in the TDL as it is. Renaming constants should in fact also involve indentation/space/linebreak cleanup, which is a separate task. (And to continue hinting at something I think about this issue generally: I do not believe there is any hope for bringing a project like the Matrix to a consistent state, ever. Why? Because in order for sucn initiatives to lead, on average, to anything apart from lengthy discussions, a military-grade authority is needed, which I don't think will ever be present in the project, nor would any of us want it to be, I don't think. You might think right now: oh but it is Olga who is arguing over this, but trust me, if we try to institute any kind of protocol, you will soon find yourself arguing over something you didn't even expect you would.)

olzama:
Thanks a lot, @dantiston and @goodmami! 
Interesting discussion :).

gmcs/constants.py
goodmami: One thing at a time :)

Regarding consistency, this is why PEP-8 and projects like [Black](https://github.com/psf/black) exist. Just defer to those rules and remove the question. But there's still a bit to do before we can get there.

gmcs/linglib/lexical_items.py
goodmami: I can't agree with this. We're at an impasse.

gmcs/linglib/wh_ques.py
olzama: Oh I see, right.

gmcs/constants.py
olzama: > this is why PEP-8 and projects like [Black](https://github.com/psf/black) exist

Oh, I know they exist but I still believe what I said. In order for a project to actually follow any guidelines in practice, a military-grade authority is needed within the project. Regardless of whether the guide came from. So our disagreement lies in you thinking we can get there and me thinking we cannot. I don't claim to actually know who is right of course :).

dantiston: I think a major piece of infrastructure that is missing right now (that we should probably make an issue for since it’s not there yet) is PR hooks and pre-commit hooks. The way these things are managed at my work very well across Python, C++, PHP, JavaScript, and more, is when you submit a PR, the server tells you if you’re breaking the rules. You *can* break the rules if you get a reviewer to accept, but there’s lots of signals about style problems, test failures, typing problems, etc. Yes, they aren’t 100%, but I think Black’s philosophy about no-config reformatting is great because it removes a category of disagreement. 

There’s of course a balance to find between writing styling rules and using them, but moving towards automated tooling and commit-blocking checks is exactly the “military-grade authority” that will allow an async project like this to maintain style over time.

Last note: I have seen people write linter rules to complain about invalid constant names, so it’s possible if we want to do it :-) (and there‘s probably something existing we could use)

olzama: I don't seem to be able to get my "military-grade authority" point across. What I mean is, it can only come from a person who has actual power over the project. Such as the project owner. This project has always emphatically exercised a very different philosophy. 

People will always have different opinions on any particular style issue. There is nothing inherently right or wrong in any particular style guide. Subscribing to one is either a personal decision, in which case it does not make sense to bring that discussion into a collaborative project at all because it can only lead to endless arguing (during which real bugs will be overlooked), or it is a project-wide decision, but that is only possible if one person who is actually the owner and has actual power over everyone else (as in: if you don't agree, I fire you) ultimately imposes the decision on everybody. Maybe something like this will happen in the Matrix one day but for now I don't see it.

olzama: That said, I personally wouldn't mind a tool which I could **optionally** run to get an idea about the degree to which my code complies with some style guide. I would probably run it and would probably even follow (some of) its suggestions. I would be very against anything that actually prevents commits though, in the context of this project. (Not that I have that military-grade say, either!)

olzama: I want to note that we already seem to be spending more time arguing over style issues (in a meta-way) than, say, discussing analyses or other contentful aspects of the work. I realize it may seem that I am the instigator but (i) actually, I am not :) I push back on these things but I never initiate them; and (ii) just wait till a project member comes on board who has strong views on style which are different from Mike's and T.J.'s Or even an issue surfaces on which Mike and T.J. happen to disagree. 

No no, I really am of the opinion that style should be kept firmly at bay and, come to think of it, not even come into CR (because it simply is too distracting).

olzama: To conclude my above thoughts, I would be for encouraging using an automated tool which highlights style locally according to some guide (PEP or whatever) and for more or less banning (refraining from discussing) style in code reviews.

olzama:
Re style policies: given that we seem to fundamentally disagree, maybe we should have a zoom meeting some time, instead of discussing it further via the forum? I think the forum both (i) can come across not the way the person means, and (ii) actually brings out additional aggressiveness/defensiveness in people sometimes. So maybe a meeting would be preferable (for communication purposes rather than for making any specific decisions, if for no other reason than simply because it is not very clear how any decision could technically be made :) ).

goodmami:
Sounds good. Let's set it up by email.

PR #511, opened 2020-09-01
Wh cleanup

### LINKED ISSUE #510: WH feature needs to be simplified
The WH feature does not need to be as complex as it is now. At a minimum, it should only contain an OR and no AND. At a maximum, if the overall analysis of wh-questions is simplified after being integrated with information structure, it could just become a boolean feature.

olzama:
The WH feature was unnecessarily complex; I was not using the AND portion of it, and it is not very likely it will be needed in the future (at least no indication of it right now).

So it is now of type *logical-or*.

Some time in the future, it may even just become a *bool* and only serve question particles (if the analysis of wh-questions becomes simpler and it is no longer needed in embedded questions) but not yet.

### DIFF
--- a/gmcs/linglib/clausalcomps.py
+++ b/gmcs/linglib/clausalcomps.py
@@ -633,7 +633,7 @@ def customize_clausal_verb(clausalverb,mylang,ch,cs,extra):
 
     if not is_nominalized_complement(cs):
         if cs['ques'] == 'prop':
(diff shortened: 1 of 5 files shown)


olzama:
I will merge these without review; I need to get going on other stuff that depends on this, and I am starting to get confused with all my branches... This one is really small and simple.

PR #512, opened 2020-09-01
0 alist cleanup

### LINKED ISSUE #488: Get rid of 0-alist, 0-1-alist, etc, in the customization system.
Use explicit syntax like ICONS.LIST < > instead (for 0-alist).

_Originally posted by @olzama in https://github.com/delph-in/matrix/pull/482#issuecomment-666562157_

olzama:
This is to address #488. Guy says those types should not be used because they "lead to bugs". I can't say I fully understand that but at any rate, the code was inconsistent between 0-alist vs LIST < >, so, had to make it consistent anyway and will go with Guy's recommendation.


olzama:
I will merge these without review; I need to get going on other stuff that depends on this, and I am starting to get confused with all my branches... This one is really small and simple, except I merged the other one into it and then accidentally pushed... Primarily decided to go ahead and merge because of that.

PR #516, opened 2020-09-10
Some comments in matrix.tdl to be removed

olzama:
I was doing some unrelated work in matrix.tdl and came across this. Perhaps this can be removed at this point.

### DIFF
--- a/matrix-core/matrix.tdl
+++ b/matrix-core/matrix.tdl
@@ -1670,19 +1670,9 @@ binary-rule-left-to-right := basic-binary-phrase &
 binary-rule-right-to-left := basic-binary-phrase &
   [ ARGS < [ KEY-ARG bool ], [ KEY-ARG + ] > ].
 
-; ERB 2005-05-04 If messages are introduced by unary rules at the
-; top of each clause, most phrase types should say [MSG no-msg]
-; on the mother.  The exceptions to this are the clauses themselves
-; and coordination constructions.  Basic-head-only is for
-; the clauses.
-
-; ERB 2007-01-21 Removing no-msg-phrase.
-
 head-only := unary-phrase & headed-phrase &
   [ HEAD-DTR #head,
     ARGS < #head > ].
-
-;head-only := basic-head-only & no-msg-phrase.
 
 head-initial := binary-headed-phrase &
   [ HEAD-DTR #head,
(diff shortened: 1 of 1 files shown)


dantiston (approved):

goodmami:
I'm quite happy to cut off these vestigial comments, but I'm not sure what method you use to identify them. E.g., even your recent comment about `clause` looks like those that are removed, except that it is newer:

```tdl
; OZ 2020-01-25 Removing the QUE constraint to allow in-situ questions.
clause := phrasal.
```

Anyway, last time I asked Emily about removing these kinds of comments (years ago) she was opposed, so it might be worth running this one by her before merging.

dantiston:
As I mentioned last time we discussed this I think, I can see lots of value in comments *about existing code and former code*, like this:
```
# a is a function of x
# a = f(x)
# nope, a is a function of x and y
# a = f(x, y)
# okay, language is hard
a = f(x, y, z)
```

But when the final result is `this code got moved to xyz.py` or `don't need a anymore`, I think the comments simply become distractions!

I think the ones @olzama has removed fall into the latter category and the remaining (in the files changed preview) fall into the former.

goodmami:
Hmm, I see. In that case, having these comments more descriptive might be better, e.g.:

```
; OZ 2020-01-25 Removing the 'QUE.LIST < >' constraint to allow in-situ questions.
```

Because otherwise "the QUE constraint" is a bit vague. But I'm not suggesting that we go back and clarify all the comments.

olzama:
Right; this was a bit of an experiment on my part. I was working on something else and just came across the comments which it seemed to me are no longer informative. There was no other method; just coming across them at random. What I wanted to experiment with was jumping back and forth between the trunk and my branch, and suggesting changes not in the branch which I am working in, but a new one, created specifically for this thing that I just randomly came across, which has nothing to do with my current task.

Now I am going to try and request a review from @emilymbender :) Emily, if you find the interface too frustrating please don't torture yourself :). But otherwise, if you click on Files Changed:

Screen Shot 2020-09-10 at 8 28 18 PM

See if what it shows to you is informative? There isn't much. And then you can either leave a post here with your opinion or there is, I believe, an option to comment on particular lines, by clicking the plus sign, which you get by hovering over the stuff:

Screen Shot 2020-09-10 at 8 29 49 PM

emilymbender:
Thanks to your pointer, Olga, I was able to see the diffs. In general, I think it is good to keep old comments around so long as they are connected in some way to current analyses. I believe the ones you have identified are now completely irrelevant, but I agree with @goodmami that the comment on the change about QUE should be made slightly more verbose, since you are removing the previous version of the code that showed what it looked like.

olzama:
Thanks, Emily! I will edit the comment.

Issue #517, opened 2020-09-14
web questionnaire for wh-inflection


olzama:
The current web questionnaire for wh inflection shows too many choices:

<img width="767" alt="Screen Shot 2020-09-14 at 1 30 47 PM" src="https://user-images.githubusercontent.com/10963114/93135068-92fc0400-f68e-11ea-8b10-ac36c3f480f8.png">

Whereas the only good result that is currently possible is just:

section=wh-q
wh-q-infl=on

1. The questionnaire should only show valid choices
2. To reflect the current behavior of the back end, the choices should probably be "subject or object" (meaning either one or both), and nothing else.
3. I suspect valence change operations should not be available for this one, at least not now.

Now, it is possible that (2) should be changed along with the back end (the verb should probably care, which element is wh, if in the future we will be linking things properly in the MRS), but that's a different issue.

emilymbender:
added object seems to be related to the valence change library and higher-ranked/lower-ranked to direct-inverse. It seems like added object should only be relevant if this rule is *also* doing valence change and higher-/lower-ranked only if direct-inverse is on. (At which point maybe some language does mark wh on higher-ranked argument??)

PR #518, opened 2020-09-14
Wh questionnaire

olzama:
I managed to not include the updated matrixdef when I merged wh-questions in. (I was sure I did include it!)

There is a number of things I need to change about it, as usual, and I will probably need to merge it in before all of that is done, but all feedback is very welcome. 

One thing I am working on right now is https://github.com/delph-in/matrix/issues/517.

### DIFF
--- a/web/matrixdef
+++ b/web/matrixdef
@@ -2544,8 +2544,6 @@ BeginIter qverb{i} "a Question Verb Type" 0 0
   EndIter feat
 
   Label "<p>There is no predefined case frame for question verbs. Please specify the case for the subject above.</p>"
-  #Select valence "Question Verb {i} case frame" "<br /><br />Case frame: " " (only S/A case will be used)<br />"
-  #fillverbpat
 
   Radio predtype "Question Verb {i} stem {j} second predicate type" "<p>The second
   predicate that this verb contributes is <br>" "</p>"
--- a/web/matrixdef
+++ b/web/matrixdef
@@ -1736,22 +1736,6 @@ Radio front-matrix "Fronting" "<p> Question phrases can appear at the left edge 
    but there is currently no distinction between echo and canonical qiestions
    in the semantic representation that you obtain from this library.</p>"
 
-#  . nondlinked-oblig "Obligatory Multiple Fronting" ""  "of all question pronouns and adverbs
-#  (but not phrases with determiners)<br><br>"
-
-#Label "<p>Note: If a <i>wh</i>-adverb intervenes between the subject and the verb, such adverb
-#is analyzed as <i>in situ</i> and will be ruled out by the obligatory multiple fronting option.</p>"
-
-    #Label "<p><b>Pied-piping:</b>
-    #In many languages, not only the question word but an entire question phrase
-    #can or must be fronted. Languages place different constraints on how this works. In English, for example:</p>"
-
-    #Label "<p><i>Which book did you read?</i><br>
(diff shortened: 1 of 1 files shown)


web/matrixdef
```diff
@@ -2764,12 +3010,14 @@ BeginIter adp{i} "an Adposition" 1
...
 
+
+
 Section morphology "Morphology"
 
 Cache nouns noun[0-9]+$ name
 Cache verbs verb[0-9]+$ name
+Cache qverbs qverb[0-9]+$ name
 Cache auxes aux[0-9]+$ name
-Cache dets det[0-9]+$ name
```
dantiston, line 2772: Was this not being used?

web/matrixdef
```diff
@@ -3382,6 +3636,9 @@ BeginIter det-pc{i} "a Position Class" 1
       MultiSelect value "Determiner Position Class {i} Lexical Rule Type {j} Feature {k} Value" "Value: " ""
       fillvalues p=det-pc{i}_lrt{j}_feat{k}_name
 
+      Select head "Determiner Position Class {i} Lexical Rule Type {j} Feature {k} Head" "Specified on: " "" "" "toggle_element.call(this, 'adj-pc{i}_lrt{j}_adj_pc_note', 'on', ['subj', 'mod'])"
```
dantiston, line 3640: I think you don't want this `toggle_element.call(this, 'adj-pc{i}_lrt{j}_adj_pc_note', 'on', ['subj', 'mod'])` call here, that's for showing the switching type LRTs message when defining a switching type adjective.

dantiston: Also, I'm not sure why we want a selectable that only has one option at all?

dantiston (approved):
Couple small changes, otherwise looks good! Congrats!

web/matrixdef
olzama: Hmm. This does not ring a bell. I remember removing something that used to be called "qdet" but not "det". (I wish we could have a few automated tests for the questionnaire... right now it is tricky because the website often isn't even working :) ).

olzama: Thanks for this, T.J., this should of course stay.

web/matrixdef
olzama: Right! We don't. I think what happened is I couldn't, at the time, figure out how to do it differently and decided to simply copy what was working :). I will see if I can quickly figure it out now, and if not will file an issue for the future...

olzama: OK, thanks for pointing this out! I can't test this at all because the website simply doesn't work with python3 version. I can test it separately in the python2 version but I think it's best to leave it till the time I can properly test it in this version.

olzama:
I am unable to test this new questionnaire on the web, even with the python2 version of the code. Any new sandbox isn't working (https://github.com/delph-in/matrix/issues/521). I wonder if I should not merge this for now? I didn't really test it, to be honest. It should be similar to what I currently have working in my sandbox but there are slight differences, and, it being the questionnaire, there is hardly a guarantee that the diffs aren't in fact breaking it. And then there is the risk of the questionnaire being broken in the trunk. On the other hand, currently there is no wh questionnaire in the trunk at all. P.S.: I am afraid to even touch the sandbox; it is the last version I have left that's working at all :).

olzama:
I guess the diffs (between this and the version that I have in the only sandbox that I have working) are really almost nonexistent (just a few comments), so I will go ahead and merge it... fingers crossed.

PR #519, opened 2020-09-14
Remove QUE constraint on clause and update on wh-test

olzama:
This came up as part of my "get rid of lexical threading" work item but I think it is kind of separate (at least it certainly doesn't require getting rid of lexical threading first), and it involves updating a test, so I made a separate PR.

The description is in the commit messages (hopefully detailed enough).


gmcs/linglib/wh_ques.py
```diff
@@ -133,10 +133,13 @@ def customize_wh_ques(mylang, ch, rules, roots):
...
     if ch.get(NO_MULTI) == ON:
-        mylang.add('''clause :+ [ SYNSEM.NON-LOCAL.QUE.LIST 0-1-list ]. ''')
-        if ch.get(MTRX_FRONT) in [SINGLE]:
+        if ch.get(MTRX_FRONT) == SINGLE:
             mylang.add(
                 '''wh-ques-phrase := [ HEAD-DTR.SYNSEM.NON-LOCAL.QUE.LIST < > ].''')
+        if ch.get(MTRX_FRONT) == IN_SITU or \
+                (ch.get(MTRX_FRONT) == SINGLE and ch.get(MTRX_FR_OPT) == NONE_OBLIG):
+            mylang.add(
+                '''insitu-int-cl := [ HEAD-DTR.SYNSEM.NON-LOCAL.QUE.LIST < [ ] > ].''')
```
dantiston, line 142: Naïve syntax question. Does this `< [ ] >` means non-empty list or empty list?

dantiston (approved):
lgtm

gmcs/linglib/wh_ques.py
goodmami: non-empty, and specifically one item. Empty and terminated is `< >`. Empty and open is `< ... >`. (an open list can be unified with another list). The `[ ]` just means that there are no constraints (an empty AVM) on the thing that is there.

Also see here: http://moin.delph-in.net/TdlRfc#Cons_Lists

I'm not sure if the APPEND lists do anything special here, though.

emilymbender: I don't think there's anything different in this respect because of the append list structure. The value of LIST there is still a list, and < > et al are still interpreted in the same way.

PR #523, opened 2020-09-18
Fix 522

### LINKED ISSUE #522: rtest doesn't ignore .DS_store
I am not sure this should be dealt with at the level of rtest, or perhaps it is my own responsibility to make sure .DS_store doesn't appear where I don't want it? But I don't know how to do that, and I am sure most people don't either. Could we make it so rtest --clean deletes .DS_store or/and rtest doesn't try to treat it as a test if it is there?

I realize that .DS_store being a OSX thing might make dealing with it on the level of rtest unappealing.

goodmami:
This does 3 things:

- Expands the `.gitignore` files. This is in fact orthogonal to #522 but it falls under the general topic of better ignoring non-test files.
- Makes it easier to add filenames that should be ignored when discovering tests. There is now a module variable `IGNORE_FILES` for this.
- Add `.DS_Store` to the set of files to ignore.

Fixes #522

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,14 +1,9 @@
+### Matrix-specific ###
+
 # Ignore cache, testing, build, and other temporary directories
 tests/regression/grammars/
 tests/regression/logs/
 tests/regression/home/current/
-
-# Ignore the compiled python files
-*.pyc
-__pycache__/
-
-# Ignore auto-saves
-*~
 
 # Ignore files under home/gold/testname which are not needed
 tests/regression/home/gold/*/update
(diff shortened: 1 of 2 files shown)


dantiston (approved):
LGTM

olzama:
This all sounds fine---although actually, it reminds me that I was keeping .DS_Store in my global ignore list because I had once read this: https://julien.danjou.info/properly-managing-your-gitignore/

But, I don't particularly care. The rtest changes certainly look good! And like I said, if we are adding OSX-specific stuff to rtest (which I myself requested), might as well add it to gitignore.

olzama (approved):

goodmami:
Thanks for the reviews

> it reminds me that I was keeping .DS_Store in my global ignore list because I had once read this: https://julien.danjou.info/properly-managing-your-gitignore/

I admit I don't have quite as well-honed an opinion regarding `.gitignore` as I do for other parts of the codebase, but the link makes sense. Most of the stuff I added here are from the default set that GitHub provides, with just a handful (`.DS_Store`, autosave files, etc.) added. Even with the default Python stuff there's a bunch I doubt we'll ever use and they could be weeded out. But I guess let's go with what we have here and remove some later if they become a problem.

PR #524, opened 2020-09-25
Different treatment of interrogative morphology

olzama:
This is another PR which is preparation for getting rid of lexical threading. Without lexical threading, morphological marking of questions needs to be done differently; we can no longer just say that the verb's QUE list is empty or not, and call it a day. 

There are only 7 actual files to review here (the rest is two regression tests which I deleted).

As part of these changes, I got rid of the option for wh-only marking (separate from polar), at least for now. There aren't any languages which do that, to my knowledge (I thought Yukaghir was but I was mistaken), and the analysis for that was very unclear to me. What I had "working" was convoluted and probably wouldn't have worked in the long run. This analysis can be added back at some point when I am more sure about it.

For now:

- [x] Removed regression tests for wh-only morphology
- [x] Removed the wh-only option from the Questionnaire; renamed "both" "polar and wh"
- [x] Added new property for Lexical Rule Types: "interrogative" (in parallel with "evidential" and others)
- [x] Added method write_interrogative_rules() (in parallel with "write_evidential_behavior()" etc.; disliked the term "behavior" in this context and opted for less consistency, but could be convinced otherwise. More importantly, the method has a flaw: it allows adding both polar-only and polar-and-wh rules to the same position class. That will create ambiguity because polar-and-wh will fire where polar-only will as well (if the orthography is the same?). But it is not very likely that the user will specify rules like this (and with the same orthography, too?), and the alternative seems for now most convoluted: two wh-only rules added instead of the user-specified rule, with different names, which necessitates multiple changes on many levels of the morphotactics library, unless I find a simpler way. I will try to revisit this later and will file an issue.
- [x] Deleted the code in features.py which was previously making the rules interrogative. This work is now done by morphotactics.py (which I think is semantically better, as "question" is a pseudofeature, not a proper feature. 
- [x] Skip the two information structure lexical rules tests which I had added experimentally before, as they no longer work with this new treatment of question morphology , and I don't understand why and couldn't figure it out after some debugging. Getting rid of lexical threading is of higher priority, because information structure lexical rules never actually worked (at least it seems so: https://github.com/delph-in/matrix/issues/494). Comment out related code with clarifying comments.
- [x] A new parametrized list is now needed, *non-wh-list*, which means it is a list of QUE-nonempty things. It is needed for lexical rules which require that the COMPS list is either empty or does not contain a wh-word.


olzama:
@goodmami (of course only if you can) Asking for your review as you know quite a bit about the morphotactics library. Do you have an opinion on turning a user-specified LRT into two different ones? How should this be done? (See the description of the issue above and in *write_interrogative_rules()* in morphotactics.py)

@emilymbender -- I wonder if the *write_interrogative_rules()* in morphotactics.py, if you look at it, gives you more clarity about the issue than I was able to provide with [the Discourse post](https://delphinqa.ling.washington.edu/t/no-lexical-threading-and-morphological-marking-of-questions/569/30)? If not that's OK;  we can revisit this later. If you have any other comments feel free to leave them here.

olzama:
I am going to merge this now so that I can continue with (no) lexical threading merge, but perhaps it could still be reviewed at some point and any issues can be addressed then.

gmcs/linglib/morphotactics.py
```diff
@@ -1047,6 +1053,48 @@ def write_pc_adj_syntactic_behavior(lrt, mylang, choices):
                 mylang.add(lrt.identifier() +
                            ''' := [ SYNSEM.LOCAL.CAT.HEAD.PRD - ].''')
 
+
+'''
+OZ-2020-09-24
+The interrogative inflection without lexical threading
+requires an interrogative supertype (which will work for
+grammars where the same morphology is used for polar and wh-questions),
+ and a further subtype which can be added if this morphology
```
dantiston, line 1062: Your comment starts to have a space at the beginning here? Small odd thing.

gmcs/linglib/morphotactics.py
```diff
@@ -1047,6 +1053,48 @@ def write_pc_adj_syntactic_behavior(lrt, mylang, choices):
...
+'''
+
+
+def write_interrogative_rules(lrt, mylang):
+    ITRG_LEX_RULE = '''itrg-lex-rule := add-only-no-ccont-rule & 
+    [ SYNSEM.LOCAL.CONT.HOOK.INDEX.SF ques ].'''
+    PROP_LEX_RULE = '''prop-lex-rule := add-only-no-ccont-rule & 
+    [ SYNSEM.LOCAL.CONT.HOOK.INDEX.SF prop ].'''
+    POLAR_LEX_RULE = '''polar-lex-rule := itrg-lex-rule & 
+    [ SYNSEM.LOCAL.CAT.VAL [ SUBJ < [ NON-LOCAL.QUE.LIST < > ] >, COMPS non-wh-list ] ].'''
```
dantiston, line 1083: Why are these variables capitalized? Did you intend to make them constants somewhere?

dantiston (commented):
Just a couple tiny style things to address in the future. Sorry for the delayed review! The other changes looked good to me!

gmcs/linglib/morphotactics.py
olzama: Hmm good question, I was thinking of them as constants because it is not like they are to be modified at any point? They are to be passed to tdl.add() as is, in various places in the code?

PR #526, opened 2020-09-25
Reshape adjunct extraction hierarchy

olzama:
The hierarchy for adjunct extraction had three supertypes which should perhaps be combined into a single one. Also, *extracted-adj-only-phrase* can in fact inherit from extracted-adj-first-phrase.

There is an [ongoing discussion on whether or not the PLACEHOLDER feature should stay](https://delphinqa.ling.washington.edu/t/finalizing-adjunct-extraction-rules-hierarchy/582/5); if we decide that it should not, that can be a separate PR. The point of this one is to clean up the existing hierarchy a bit.

### DIFF
--- a/gmcs/linglib/yes_no_questions.py
+++ b/gmcs/linglib/yes_no_questions.py
@@ -188,7 +188,7 @@ def customize_yesno_questions(mylang, ch, rules, lrules, hierarchies, roots):
                 '''non-local-none :+ [ YNQ.LIST < > ].''', section='addenda')
             mylang.add(
                 '''basic-filler-phrase :+ [ SYNSEM.NON-LOCAL.YNQ.LIST < > ].''', section='addenda')
-            mylang.add('''extracted-adj-phrase-simple :+ 
+            mylang.add('''basic-extracted-adj-phrase :+ 
             [ SYNSEM.NON-LOCAL.YNQ #ynq,
               HEAD-DTR.SYNSEM.NON-LOCAL.YNQ #ynq ].''', section='addenda')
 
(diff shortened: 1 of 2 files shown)


guyemerson:
The shape of hierarchy is quite dependent on the discussion of the PLACEHOLDER feature...

But I'm glad you've consolidated the `-super` and `basic-` types.

The `extracted-adj-only-phrase` looks broken (it doesn't use the PLACEHOLDER), but I guess that type might be going anyway.

As a more minor comment, there is inconsistent use of tabs and spaces.  I would suggest converting all tabs to spaces to avoid the code displaying badly with the wrong tab size.

olzama:
It inherits from *-first*, so, PLACHOLDER would be the same as there. But, I also checked just now and noticed that I am not using it, in fact.

olzama:
> The extracted-adj-only-phrase looks broken (it doesn't use the PLACEHOLDER), but I guess that type might be going anyway.

It inherits from _-first_, so, PLACHOLDER would be the same as there. But, I also checked just now and noticed that I am not using it, in fact.

olzama:
> As a more minor comment, there is inconsistent use of tabs and spaces.

Yes, it is all over the place in matrix.tdl and in the customization system, so, any change to either typically appears with the same style issues. I wish we had a TDL IDE...

olzama:
I think even if PLACEHOLDER does go away, it is worth having a cleaner version of something that works with lexical threading.

guyemerson:
> 
> 
> > The extracted-adj-only-phrase looks broken (it doesn't use the PLACEHOLDER), but I guess that type might be going anyway.
> 
> It inherits from _-first_, so, PLACHOLDER would be the same as there. But, I also checked just now and noticed that I am not using it, in fact.

Oh, I missed that.  These rules should be mutually exclusive (as in, the mother's SLASH list should be different for each rule), or else there would be spurious ambiguity.  On seeing the `only` rule, I assumed that it could apply when the daughter had an empty list, and the other two could apply when the daughter had a non-empty list.  But I now see that the `last` rule requires a non-empty list while the `first` rule doesn't, so they're mutually exclusive and the `first` rule subsumes the empty list case.  So as you noticed, the `only` rule isn't necessary.

PR #528, opened 2020-09-25
WH and MODIFIED were not copied up by some phrases

### LINKED ISSUE #527: Argument drop rules should propagate WH from daughter to mother

olzama:
This became apparent when I was getting rid of lexical threading, but it can be fixed separately from lexical threading.

WH should be copied up by argument-drop rules. MODIFIED should be copied by head-subject and head-complement. Otherwise after I get rid of lexical threading, regressions happen.

On the other hand, adding these constraints helps get rid of a couple of spurious parses in the Russian test (so I updated it).


dantiston (commented):
Looks good! Sorry for the belated review.

olzama:
Thanks for all the reviews, T.J.!

PR #529, opened 2020-09-28
Complementizer: minor supertype change

olzama:
A minor change which can be merged in separately from getting rid of lexical threading (which came up while I was working on the latter).

The complementizers (1) should not be constrained to be [ L-QUE - ] at the supertype level; (2) should be either wh or non wh-words, as the rest of the words.


dantiston (commented):
lgtm

PR #537, opened 2020-10-02
Head-complement and adj-head rules LIGHTness

olzama:
I want to make the following changes in order to better be able to use LIGHT for extraction ambiguity curbing (see https://delphinqa.ling.washington.edu/t/using-a-parametrized-list-for-adjunct-extraction-ambiguity/583/12?u=olzama ):

the head complement rule currently identifies the mother's LIGHT with the head daughter's HC-LIGHT. I am actually not too sure what that means, but it appears that it is only crucial for subject-auxiliary inversion and for languages with question particles which act as complementizers. In all other cases, it looks like I can assume that the mother of the head-complement rule is [ LIGHT - ]. This makes sense, too, doesn't it? 

Then I can rely on head-complement being [ LIGHT - ] in most cases to say that subjects are only extracted from [ LIGHT + ] things, and this helps me rule out some of the ambiguity in languages with flexible order of extraction.

The adj-head-phrase says it inherits its HEAD-DTR's LIGHT value, but that does not seem to affect any tests, and I'd like to take it out for similar purposes (I sometimes want to be able to say that adj-head-scopal phrase is [ LIGHT - ], e.g. for Russian where there is a negation adverb but I don't want subjects to be extracted both below and above its attachment).

Does this sound good? Please see the diffs for specifics; there are only a few lines.

### DIFF
--- a/gmcs/linglib/yes_no_questions.py
+++ b/gmcs/linglib/yes_no_questions.py
@@ -38,6 +38,12 @@ def customize_yesno_questions(mylang, ch, rules, lrules, hierarchies, roots):
 
     qinvverb = ch.get('q-inv-verb')
 
+    if ch.get('q-inv') or (ch.get('q-part') and ch.get('q-part-order') != 'second'):
+        mylang.add(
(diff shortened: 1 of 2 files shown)


olzama:
Branch mess up; please ignore this one.

PR #547, opened 2020-12-21
Modified gmcs/deffile.py to resolve some issues with getting it to ru…

bmgraves:
…n on apache.

Hard coded encoding to utf-8 when opening the matrixdef file.

Removed lambda function that was leftover from py2 version of code, replaced with suggestion from Michael Goodman.


olzama:
I merged this in but actually I couldn't make things work just yet. Installing this version on patas does not lead to a working questionnaire website. But, I may be installing incorrectly?

PR #549, opened 2020-12-28
improve mixed-case rules

lizcconrad:
Changes that reduce the number of non-inflecting lexical rules when a language has mixed case

### DIFF
--- a/gmcs/linglib/case.py
+++ b/gmcs/linglib/case.py
@@ -279,13 +279,10 @@ def customize_case_adpositions(mylang, lexicon, trigger, ch, case_pos):
 
             abbr = name_to_abbr(cn, cases)
 
-            adp_type = TDLencode(abbr + '-marker')
-            if adp_type in adp_type_names:
-                adp_type = adp_type + '_2'
-                while True:
-                    if adp_type not in adp_type_names:
-                        break
-                    adp_type = adp_type[:-1] + str(int(adp_type[-1])+1)
+            # the type name for the adp marker includes the orthography of the marker at the end
+            # this serves to ensure each marker has its own lexical entry
+            # and prevents them from being "merged" due to having identical names
+            adp_type = TDLencode(abbr + '-marker_' + orth)
             adp_type_names.append(adp_type)
 
             typedef = \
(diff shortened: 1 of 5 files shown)


gmcs/linglib/case.py
```diff
@@ -279,6 +280,14 @@ def customize_case_adpositions(mylang, lexicon, trigger, ch, case_pos):
             abbr = name_to_abbr(cn, cases)
 
             adp_type = TDLencode(abbr + '-marker')
+            if adp_type in adp_type_names:
+                adp_type = adp_type + '_2'
```
olzama: I know this is what was here before, so, not your "fault", so to speak, but this "_2" here doesn't look great. Why 2? Can there only ever been 2 asposition types? Or what does it mean. In combination with the while True...break below, the code becomes just sort of confusing overall. Given that you've debugged this, do you think you could add a comment explaining what the code is doing exactly? E.g. in terms of an example of input and output, what comes in and what the result is of this being executed.

olzama (changes requested):
Just in case you *can* provide a clarifying comment above the lines on which I remarked. If you simply restored what was there before and it would be hard for you to tell what the code is doing, I don't think it is your responsibility necessarily and the whole thing can be merged.

gmcs/linglib/case.py
lizcconrad: I updated the code so it doesn't do this anymore. What it was doing was creating a new type with "_2" at the end and then doing a loop and incrementing that integer until a type with that integer wasn't found and breaking from the loop. So it would do "dat-marker_2" then if that existed it would increment it to _3 and _4 and so on. 

But I made a change so that instead of doing this it just names the type "dat-marker_[orthography for the marker]" to ensure it's always unique. You can look at the change and tell me if approach was bad practice.

I don't know how to update the pull request though... I committed and pushed the change but how do I make the pull request reflect it?

olzama: Yes, it will! The "files changed" will reflect the most recent commit. 
Thank you so much for clearing up the code; that portion was really confusing :).

olzama (approved):

PR #555, opened 2021-01-12
Be explicit about encodings in matrix.cgi

### LINKED ISSUE #553: Upload choices file button leads to (probably encoding) error
Pick any choices file and try uploading it via the Upload Choices File button. Get the Internal Server Error. The same choices files e.g. the sample ones are loaded fine via the links.

Testing site: https://matrix.ling.washington.edu/trunkpy3/matrix.cgi

goodmami:
This commit does 3 things:
  1. Always opens files in matrix.cgi with an explicit encoding (except
     tar files and urls), currently set to utf-8-sig on read and utf-8
     on write.
  2. Decodes the bytes from uploaded files with utf-8-sig.
  3. Stores the joined path of a choices file in a variable and reuses
     it instead of recomputing it each time (unrelated to the current
     issue).

Fixes #553

### DIFF
--- a/matrix.cgi
+++ b/matrix.cgi
@@ -24,6 +24,13 @@ from gmcs.choices import ChoicesFile
 from gmcs.linglib.toolboximport import import_toolbox_lexicon
 
 
+# Sometimes UTF-8 files have a (gratuitous) BOM. The utf-8-sig
+# encoding will strip the BOM, but we want to always write files
+# without it, so use regular utf-8 on write.
+READ_ENCODING = 'utf-8-sig'
+WRITE_ENCODING = 'utf-8'
+
+
 cgitb.enable()
 
 # Production Check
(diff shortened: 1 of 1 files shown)


dantiston (approved):

olzama:
Thank you very much for helping us out with this, @goodmami ! It appears that at least the immediate issue we were observing is fixed with this commit.

Testsite: https://matrix.ling.washington.edu/cgifix/matrix.cgi

I do not have the energy to test much right now, but at the very least I am able to upload the Japanese sample choices file (using the upload button).

This is very promising!!

olzama:
Tagging @bmgraves  just in case, so he doesn't miss this latest development and doesn't continue debugging on his own Confirmed also that a random choices file from the regression tests loads.

goodmami:
Thanks @olzama and @dantiston! I've switched this from a draft to a regular PR so it can be merged.

I want to add that this is very much a duct-tape kind of commit. For the continued health of the project that matrix.cgi file should really be replaced entirely with something more standard, like a WSGI app or a split RPC-backend and Javascript frontend, as we've discussed. But for now I'm glad it seems to have helped.

Issue #622, opened 2022-01-28
Validation error in adnominal possession


emilymbender:
Validation should block giving two separate values for person on a possessive pronoun. (Also test whether similar issues occur elsewhere.)
[choices-per.txt](https://github.com/delph-in/matrix/files/7962250/choices-per.txt)

ltxom:
Could you please check my [pull request #641](https://github.com/delph-in/matrix/pull/641) for this fix?

PR #641, opened 2022-04-28
Fix Issue #622

ltxom:
As validate.py failed to block giving two separate values for the person on a possessive pronoun, the solution creates `duplicated_person_feat` boolean variable to test if there're two person features. If yes, it will show an error `You must only give one person value for this feature` at the index of duplicated person feature.

### DIFF
--- a/gmcs/validate.py
+++ b/gmcs/validate.py
@@ -1759,11 +1759,9 @@ def validate_adnominal_possession(ch, vr):
                     mess = 'You must give the spelling for this pronoun.'
                     vr.err(inst.full_key+'_orth', mess)
 
-                # These variables are used to test if there are duplicated person features
+                # These variables are used to test if there are duplicated features
                 # in poss-pron_instance_feat path
-                has_person_feat = False
-                duplicated_person_feat = False
-                duplicated_person_index = -1
+                features_map = {}
                 index = 1
                 for feat in inst.get('feat'):
                     if not feat.get('name'):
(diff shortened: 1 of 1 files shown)


emilymbender:
Thanks, Tom. I'm a little surprised that this test is specific to person features. This seems like something that applies to all features.

Also, it might be helpful for the error message to say something like: If you would like this value to be a disjunctive value, please use the multi-select functionality (if available).

ltxom:
Got it! I will modify it to test all duplicated features and change the error message.

ltxom:
I just updated it. Now it can check all duplicated features. For example:
```
section=adnom-poss
  poss-pron1_type=non-affix
  poss-pron1_agr=non-agree
  poss-pron1_mod-spec=spec
  poss-pron1_order=head-final
    poss-pron1_instance1_name=bug
    poss-pron1_instance1_orth=my
      poss-pron1_instance1_feat1_name=person
      poss-pron1_instance1_feat1_value=1st
      poss-pron1_instance1_feat2_name=person
      poss-pron1_instance1_feat2_value=2nd
      poss-pron1_instance1_feat3_name=case
      poss-pron1_instance1_feat3_value=none
      poss-pron1_instance1_feat4_name=case
      poss-pron1_instance1_feat4_value=none
  poss-pron1_possessum-mark=no

./matrix.py v choice:
poss-pron1_instance1_feat2_value
   If you would like this value to be a disjunctive value, please use the multi-select functionality.
poss-pron1_instance1_feat4_value
   If you would like this value to be a disjunctive value, please use the multi-select functionality.
```

gmcs/validate.py
```diff
@@ -1758,13 +1758,30 @@ def validate_adnominal_possession(ch, vr):
                 if not inst.get('orth'):
                     mess = 'You must give the spelling for this pronoun.'
                     vr.err(inst.full_key+'_orth', mess)
+
+                # These variables are used to test if there are duplicated features
+                # in poss-pron_instance_feat path
```
olzama: Is this part of the comment ("in poss-pron_instance_feat path") still relevant?

gmcs/validate.py
```diff
@@ -1758,13 +1758,30 @@ def validate_adnominal_possession(ch, vr):
...
+                features_map = {}
+                index = 1
```
olzama, line 1774: I would add a comment explaining what the role of `index` is versus `count`

ltxom: Hi Olga, thank you for your suggestions! Please check my updated comment.

olzama: Ah OK. This key retrieval has always been rather confusing :/. But I think I now remember.

Issue #621, opened 2022-01-28
Adnominal possession library doesn't fully create disjunctive case types


emilymbender:
The attached choices file leads to a grammar that won't load. It seems that the adnominal possession library knows it should invoke disjunctive types (here non-abs, for all case values except abs), but doesn't actually create them
[bug-choices.txt](https://github.com/delph-in/matrix/files/7961899/bug-choices.txt)
.

PR #642, opened 2022-05-07
Fix issue #621

ltxom:
Here are my findings and solutions in this PR:
1. When I create lexical entries for the verbs with a disjunctive case value for one of their arguments, Matrix can save the disjunctive case into TDLHierarchy object and myland.tdl properly. However, in the adnominal possession library (as in issue #621 ), Matrix saves the case type into TDLHierarchy object but not into myland.tdl.

This is because Matrix saves case hierarchy into mylang.tdl too soon in [case.py](https://github.com/ltxom/matrix/commit/7d05de93827afeb4bc4281f59a77f7c0a113192c#diff-ca3f9b431ed3c0f7abbdfb6096e0d9757a2ce65b84fe3843a0dde5f2f665a7ebL349), it should wait until the adnominal possession library [is loaded](https://github.com/ltxom/matrix/commit/7d05de93827afeb4bc4281f59a77f7c0a113192c#diff-27e7a8c926a9cec63e6f5ee569daf640a14afe232f4920a20adc6f61d2f0290dR538-R546).

2. In [tdlhierarchy.py](https://github.com/ltxom/matrix/commit/7d05de93827afeb4bc4281f59a77f7c0a113192c#diff-63c93581c2158184c412857374f1275cf01c04d98c4a705bff08cd71fc631d96L158), the supers case value is always [self.name], which is "case." 

However, the "real-case" can be created as an intermediate case type for all real cases for possessive strategy. [If the "real-case" exists, it should be the supertype (instead of "case") of the new_type.](https://github.com/ltxom/matrix/commit/7d05de93827afeb4bc4281f59a77f7c0a113192c#diff-63c93581c2158184c412857374f1275cf01c04d98c4a705bff08cd71fc631d96R158-R164)

3. If "real-case" is necessary, Matrix can [create it multiple times](https://github.com/ltxom/matrix/commit/7d05de93827afeb4bc4281f59a77f7c0a113192c#diff-ca3f9b431ed3c0f7abbdfb6096e0d9757a2ce65b84fe3843a0dde5f2f665a7ebL117-L120) and save them into TDLHierarchy object., which has been simplified by [only saving one "real-case", "case" relation](https://github.com/ltxom/matrix/commit/7d05de93827afeb4bc4281f59a77f7c0a113192c#diff-ca3f9b431ed3c0f7abbdfb6096e0d9757a2ce65b84fe3843a0dde5f2f665a7ebR117-R120).

4. I also created a new regression test named "adnom-poss-disjunctive-case," the choice of it comes from Emily's post in issue #621. Before this PR, it failed to compile (because of the missing non-abs disjunctive case in mylang.tdl), and it can compile and parse [test sentences](https://github.com/ltxom/matrix/commit/7d05de93827afeb4bc4281f59a77f7c0a113192c#diff-91127789d27730574d5df3a986c3bd5454512b06af3384bdf1aed45602037e87R1-R9) after these fixes.

5. All other regression tests are passed after these fixes:
```
************* SUMMARY *************
Passed  517/528 tests
Failed    0/528 tests
Errors    0/528 tests
Skipped  11/528 tests (run rtest.py --list --skipped --verbose for more info)
```


gmcs/linglib/case.py
```diff
@@ -345,8 +347,13 @@ def customize_case(mylang, ch, hierarchies):
...
         hierarchies['case'].get_type_covering(adp_cases)
 
+    # LTX 2022-04-22: It is possible to generate disjunctive case from
+    # adnom_poss library (see issue issue #621).
+    # Therefore, it should save case hierarchy to mylang.tdl after
+    # adnominal possession customization.
+
     # now output the case hierarchies
-    customize_case_type(mylang, hierarchies)
+    # customize_case_type(mylang, hierarchies)
```
olzama, line 356: why not just delete the line (and I suppose the comment as well)?

olzama:
The code changes look clear!

emilymbender:
Thank you, @ltxom --- I'm merging this now.

Issue #627, opened 2022-02-03
Validate for verb types & supertypes both indicating valence.


emilymbender:
The attached choices file gives a grammar that won't load into the LKB because of a redundancy in the type hierarchy:

Redundancy involving VERB3-VERB-LEX
While evaluating the form starting at line 41, column 0
 of VERB3-VERB-LEX: INTRANSITIVE-VERB-LEX is redundant - it is an ancestor of VERB1-VERB-LEX o

Easiest fix is probably to validate for this scenario:

```
  verb1_valence=intrans
    verb1_stem1_orth=iv
    verb1_stem1_pred=_iv_v_rel
  verb2_valence=trans
    verb2_stem1_orth=tv
    verb2_stem1_pred=_tv_v_rel
  verb3_supertypes=verb1
  verb3_valence=intrans
    verb3_stem1_orth=test
    verb3_stem1_pred=_test_v_rel
```

ltxom:
Here's my findings to this issue:

The problem is that verb3 has redundant valence value 'intrans,' which leads to an incorrect TDL as:

```
verb1-verb-lex := intransitive-verb-lex.
verb3-verb-lex := verb1-verb-lex & intransitive-verb-lex.
```

The problem will be the same if the inheritance has one or more inter-mediate instances as below:

```
  verb1_valence=intrans
    verb1_stem1_orth=iv
    verb1_stem1_pred=_iv_v_rel
  verb3_supertypes=verb1
    verb3_stem1_orth=test
    verb3_stem1_pred=_test_v_rel
  verb4_supertypes=verb3
  verb4_valence=intrans
    verb3_stem1_orth=test2
    verb3_stem1_pred=_test2_v_rel

TDL:
verb1-verb-lex := intransitive-verb-lex.
verb3-verb-lex := verb1-verb-lex.
verb4-verb-lex := verb3-verb-lex & intransitive-verb-lex.
```

I want to consult a question before working on it:
Since only verbs have argument structures (valence), can Matrix create other POS (noun, adv, det) that exhibit similar issues with inheritance hierarchy? I guess there could be something to do with CASE, something like:

```
noun1-noun-lex := some-case-noun-lex.
noun2-noun-lex := noun1-noun-lex & some-case-noun-lex.
```

ltxom:
Update:

I just tried the scenario in previous reply with:

```
    noun1_feat1_name=case
    noun1_feat1_value=erg
  noun1_det=imp
    noun1_stem1_orth=n1
[...]

Issue #452, opened 2020-05-19
Treat Coordinators/conjunctions more like Lexical Items


goodmami:
**Migrated from Trac:**

> - Original: <http://lemur.ling.washington.edu/trac/matrix/ticket/875>
> - Reporter: kphowell
> - Created: 01/30/18 13:06:54
> - Updated: 01/30/18 13:06:54
> - Keywords: 
> - Notes:
>   - There may be formatting problems from the conversion
>   - Comments on tickets are *not* migrated! See the original issue.

Currently conjuctions aren't checked for name-space collisions (which is possible eg. in Mandarin). They should be added to the list of lexical types and the choices file should be set up such that they are checked by insert_ids() in lexical_items.py

Issue #445, opened 2020-05-19
Auxiliary Features Dropdown


goodmami:
**Migrated from Trac:**

> - Original: <http://lemur.ling.washington.edu/trac/matrix/ticket/863>
> - Reporter: kphowell
> - Created: 01/12/18 13:56:10
> - Updated: 01/12/18 13:56:10
> - Keywords: 
> - Notes:
>   - There may be formatting problems from the conversion
>   - Comments on tickets are *not* migrated! See the original issue.

The Auxiliary features dropdown only populates with verbal features, so it is not possible to specify features on the subject. Emily believes this should be possible.

ajcassell:
This is no longer an issue.
<img width="753" alt="Screen Shot 2024-10-03 at 6 54 47 PM" src="https://github.com/user-attachments/assets/685978f5-174e-4d9d-977c-0af8a3053b50">

PR #647, opened 2022-06-09
Fix 627

ltxom:
This PR fixes issue #627 .

There are two improvements:
1. validation reports an error if a verb inherits another verb and has valence. ([lexicon.py line #445-#452](https://github.com/delph-in/matrix/compare/trunk...ltxom:fix-627#diff-73ef70b569a1fbfbf6a4ff518128e11e38ccb20d2db78509f4a6a358d8fa9491R445-R452))
2. validation shows a question mark to warn the user of other duplicated features (maybe different values) with inheritance hierarchy.

For the second improvement, there are [existing codes](https://github.com/delph-in/matrix/blob/trunk/gmcs/linglib/lexicon.py#L259-L262) that are implemented for nouns, but this functionality is tangled with only 'det' and 'noun' POS.

Therefore, I created [a new section](https://github.com/delph-in/matrix/compare/trunk...ltxom:fix-627#diff-73ef70b569a1fbfbf6a4ff518128e11e38ccb20d2db78509f4a6a358d8fa9491R1088-R1113) and [a helper method](https://github.com/delph-in/matrix/compare/trunk...ltxom:fix-627#diff-a52fbfbe26e85fa86a3b0ccde9e13c092b434b19fd5e9639816133205d110a57R124-R138) to check for other POS.

Regression Tests are all passed after these improvements. Here's a [choices file](https://github.com/delph-in/matrix/files/8873207/issue-627-choice.txt) to test (demo in https://matrix.ling.washington.edu/ltxom-test/)

### DIFF
--- a/gmcs/linglib/lexicon.py
+++ b/gmcs/linglib/lexicon.py
@@ -438,6 +438,14 @@ def validate_lexicon(ch, vr):
     seenTrans = False
     seenIntrans = False
 
+    # LTX 2022-05-16: validate if a verb inherits another verb
+    # and has valence, see issue #627
+    for v in ch.get('verb'):
+        if len(vtsts[v.full_key]) > 0 and vtsts[v.full_key][0] != '' and v.get('valence') != '':
+            mess = 'You must either remove the valence of a subtype verb ' \
+                   'or remove the inheritance hierarchy for this verb.'
+            vr.err(v.full_key, mess)
+
     for v in ch.get('verb'):
         st_anc = []  # used to make sure we don't have an lkb err as
         # described in comments above
--- a/gmcs/linglib/lexicon.py
+++ b/gmcs/linglib/lexicon.py
@@ -1095,8 +1095,7 @@ def validate_lexicon(ch, vr):
                     for feat in lt.get('feat'):
                         for parent_feat in parent_features_list:
                             if feat.get('name') == parent_feat.get('name'):
-                                mess = 'Please be aware that this type has a duplicated feature (' + \
-                                       lt.full_key + ': ' + feat.full_key + \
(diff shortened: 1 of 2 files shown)


emilymbender:
Thank you for this! I'd like to suggest a change to the validation message. Instead of:

mess = 'You must either remove the valence of a subtype verb ' \
                   'or remove the inheritance hierarchy for this verb.'

I propose:

mess = 'A verb class that specifies a value for valence can't also inherit a value for valence.'

Regarding the warning for the noun, it's a bit surprising that it shows up next to "add a feature". Is it possible to have the red ? show up next to the feature/value pair instead?

ltxom:
Thank you for your feedback! I updated the error message and moved the noun's warning next to the feature/value pair as you suggested.

emilymbender:
Looking better! I want to suggest a rephrasing for the warning on the nouny features. Instead of: 

mess = 'Please be aware that this type has a duplicated feature (' + \
                                       lt.full_key + ': ' + feat.full_key + \
                                       ') with inheritance hierarchy, ' \
                                       'which can be factors to create redundant or conflicting grammars.'

How about:

mess = 'A value for this feature is already specified on a supertype for this noun class. If the value specified here is conflict with that other specification, the grammar will not compile.'

The identifier of the feature doesn't need to be included in the message, because the position of the red ? makes it clear which feature is at issue.

ltxom:
Thank you for the suggestion! Please check it now :)

emilymbender:
Missing a space between "other" and "specification" but otherwise looks good!

ltxom:
Oops, my fault. It should be good now.

emilymbender:
Looks good! Merged.

PR #651, opened 2022-07-10
Fixed python list error for test by gen

rosypen:
This change fixes the test by generation page so it doesn't crash with a python error anymore.

However the generated sentences are still missing.


rosypen:
I left a comment in the forked copy of this repo (rosypen/matrix) for this commit that has more context about this change.

https://github.com/rosypen/matrix/commit/41c785de92e43a6f6bee72ef882e32bdfec8eae3#commitcomment-78115976

PR #659, opened 2022-07-22
Docstrings on a portion of lexical rule types

goodmami:


emilymbender:
Approving this, but wondering if it makes sense to refer to lexical rule types as "lexical rules". @fcbond do you have an opinion, imagining users of this documentation? I think I tend to reserve "lexical rule" for lexical rule instances.

goodmami:
Maybe something like "Type of a lexical rule that..." or "For lexical rules that..."?

emilymbender:
"For lexical rules that..." seems concise and clear.

PR #660, opened 2022-08-08
Rpendlet test by gen

rosypen:
TLDR; The basic functionality of Test By Generation remains the same but we can now see more information even when no sentences are generated.

Screen Shot 2022-08-08 at 1 41 16 AM

There are 3 commits here on this pull request I'll explain each of them.
- [Refactoring generate.py](https://github.com/delph-in/matrix/commit/de72efc8dcc41265e1b23ef74704b13ee25ff025)
This change just extracts out the logic for parsing the lkb output into its a separate function because I will eventually need to parse the ace output which has a different format entirely so I want to be able to swap out the logic for each.
- [Merge branch 'trunk' of](https://github.com/delph-in/matrix/commit/cca270587124e0fe35841187f9dc9e81b72dec08) https://github.com/rosypen/matrix [into rpendl…](https://github.com/delph-in/matrix/commit/cca270587124e0fe35841187f9dc9e81b72dec08)
Simple merge of changes in the trunk branch (which are from my last commit).
- [Populate TestByGeneration page as much as possible](https://github.com/delph-in/matrix/commit/e4732849aafeba22ca3566854bb7af8b888a7c8e)
This enables the Test by Generation page to look like the screenshot above. We can see the Patterns that were tried but still no sentences are generated. (Although I have managed to generate sentences locally, just having issues generating from the website).

This bottom photo is what the Test By Generation page looks like fully populated.

Screen Shot 2022-08-07 at 11 59 16 PM

### DIFF
--- a/gmcs/generate.py
+++ b/gmcs/generate.py
@@ -83,6 +83,12 @@ def get_sentences_from_lkb_output(output_file,  mrs_files, verb_preds):
     '''parses the output file from lkb after generating sentences to extract the sentence information.'''
     output = open(output_file, 'r')
     sentences = []
+
+     # create an entry in sentences for every mrs pattern in mrs_files
+    for i in range(len(mrs_files)):
+        sentences.append([[mrs_files[i]+":", verb_preds[i][0],
+                               verb_preds[i][1], verb_preds[i][2]], [], [], []])
+
     index = -1
     sentence = parse = mrs = state = ""
     in_output = False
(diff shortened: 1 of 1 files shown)


emilymbender:
Yay!

I'm curious -- what does "show remainder" do, if clicked?

rosypen:
It does nothing at the moment. I can investigate further but my intuition is that perhaps there already is some clipping of the generated sentences so that it isn't an explosive amount and "show remainder" will show the rest of the generated sentences when there are a lot.

emilymbender:
That makes sense. If you have time, it seems like a UI improvement would be to show that button only if there were sentences clipped.

PR #662, opened 2022-08-12
Switching Generation from LKB to ACE

rosypen:
TLDR; These changes officially switch test by generation to use ACE instead of LKB

Screen Shot 2022-08-12 at 2 18 53 PM

There are 3 commits included in this change and the titles are fairly self explanatory but there are also comments on the original commits that go into further details about what was changed.
- [switched test by gen to ace](https://github.com/rosypen/matrix/commit/5697d135243b633e0d74e7ec63226f4074a98c9d)
- [Improves preprocessing steps for generation.](https://github.com/rosypen/matrix/commit/f8978ec7e18d2949d92b3e11cdf73b0d3845dd42)
- [Removes Capitalized lexical entries sentence.](https://github.com/rosypen/matrix/commit/35154356b23671ef58a5e9f442eb7956893bc5cb)

Pending Issues
- Parse trees are still not included at this time. They will be added in the next commit
- Cosmetically, the sentences don't look 'clickable' so it would be nice to add that
- Also want to add a way to specify max sentences generated.
- A major flaw that i found while testing with other grammars is the verbs and noun selected for generation is arbitrarily the first verb/noun in the choices files. This may result in no sentences being produced if they are not compatible with each other. It may help to be able to specify the verb/nouns used in test by generation options.
- Lastly, test by generation using test sentences is also coming soon. Following adding the parse trees (:

### DIFF
--- a/gmcs/generate.py
+++ b/gmcs/generate.py
@@ -256,6 +256,12 @@ def get_sentences_from_ace_output(output_file,  mrs_files, verb_preds):
                 state = "sentence"
         else:
             continue
+    # If no sentence was added to the sentence position (sentences[x][1]) then add "#NO-SENTENCES"
+    for i in range(len(mrs_files)):
+        if len(sentences[i][1]) == 0:
+            sentences[i][1].append("#NO-SENTENCES#")
+            sentences[i][1].append("")
+            sentences[i][1].append("")
     return sentences
             
 
@@ -287,6 +293,8 @@ def get_n_predications(grammar_dir):
     choices = open(os.path.join(grammar_dir, 'choices'), 'r')
     lang = None
 
+    # Worth noting that since the stem must explicitly be [0-9] only the first 10 predicates
+    # Are ever stored for each noun type.
     pred_re = re.compile(r'noun([0-9]+)_stem[0-9]_pred')
 
     # det_re is searching for whether dets are opt, obl or imp
@@ -295,6 +303,10 @@ def get_n_predications(grammar_dir):
     # det2_re is searching for determiner predications
     det2_re = re.compile(r'det[0-9]+_stem[0-9]_pred')
 
+    # Every noun type in the choices file should have a number eg.noun23
+    # And every noun type should define exactly one determiner type (opt, obl, imp)
+    # This dictionary maps the noun number to the det type.
+    noun_det_dict = {}
     # noun_rels_dets is a list of pairs, where the first element is a noun_rel
     # the second element is whether a determiner is 'obl','opt' or 'imp'
     # for that noun_rel
@@ -302,24 +314,24 @@ def get_n_predications(grammar_dir):
     # det_list is the set of determiner rels found in the choices
     det_list = set([])
     for line in choices:
-        pline = line.lstrip().split('=')
-        m1 = pred_re.match(pline[0])
-        m2 = det_re.match(pline[0])
-        m3 = det2_re.match(pline[0])
-        if pline[0] == 'language':
-            lang = os.path.join(grammar_dir, pline[1].lower().rstrip()+'.tdl')
+        pline = line.strip().split('=')
+        if len(pline) != 2:
+            continue
+        feature, value = pline
+        m1 = pred_re.match(feature)
+        m2 = det_re.match(feature)
+        m3 = det2_re.match(feature)
+        if feature == 'language':
+            lang = os.path.join(grammar_dir, value.lower()+'.tdl')
         if m1:  # m1 is results of looking for noun predications
-            if int(m1.group(1)) <= len(noun_rels_dets):
-                noun_rels_dets[int(m1.group(1))-1][0] = pline[1].rstrip()
-            else:
-                noun_rels_dets.append([pline[1].rstrip(), None])
+            noun_number = m1.group(1)
+            if noun_number in noun_det_dict:
+                noun_rels_dets.append([value, noun_det_dict[noun_number]])
         if m2:
-            if int(m2.group(1)) <= len(noun_rels_dets):
-                noun_rels_dets[int(m2.group(1))-1][1] = pline[1].rstrip()
-            else:
-                noun_rels_dets.append([None, pline[1].rstrip()])
+            noun_number = m2.group(1)
+            noun_det_dict[noun_number] = value
         if m3:
-            det_list.add(pline[1].rstrip())
+            det_list.add(value)
 
     noun_rels_dets = remove_duplicates(noun_rels_dets)
     # print "noun_rels_dets: <<",noun_rels_dets,">><br />"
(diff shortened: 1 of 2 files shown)


gmcs/generate.py
```diff
@@ -152,6 +154,124 @@ def get_sentences_from_lkb_output(output_file,  mrs_files, verb_preds):
...
+    output.write('Compiling grammar...\n> '+compile_grammar_cmd+'\n')
+    output.flush()
+
+    ace_error = open('ace_error'+session, 'w')
+    ace_error.write('Compiling grammar...\n> '+compile_grammar_cmd+'\n')
+    ace_error.flush()
+
+    # compile the grammar
+    subprocess.run(compile_grammar_cmd.split(),
+                    stdout=output, stderr=ace_error, env=os.environ)
```
goodmami, line 175: Using `str.split()` on the `compile_grammar_cmd` can fail if there are ever spaces in the `grammar_dir` directory name. I suggest doing one of the following:

1. Put quotes around your arguments when constructing the command string, then use [shlex.split()](https://docs.python.org/3/library/shlex.html#shlex.split)
2. Just make a list instead of making then splitting a string:
   ```python
   compile_cmd_args = [
       '/usr/local/bin/ace',
       '-G', f'{grammar_dir}/{iso}.dat',
       '-g', f'{grammar_dir}/ace/config.tdl',
   ]
   ```
3. Use PyDelphin's `ace.compile()`

I think (3) is the best option because it would simplify this code a lot.

gmcs/generate.py
```diff
@@ -152,6 +154,124 @@ def get_sentences_from_lkb_output(output_file,  mrs_files, verb_preds):
...
+    ace_mrs = open('ace_mrs'+session, 'w')
+    for mrs_file in mrs_files:
+        text = collapse_mrs_to_one_line(mrs_file)
+        ace_mrs.write(text+"\n")
+    ace_mrs.flush()
+    ace_mrs.close()
+
+    # command to generate sentences
+    # TODO: add -n [count] to enumerate only the top [count] results.
+    generate_sentences_cmd = '/usr/local/bin/ace -g %(grammar_dir)s/%(iso)s.dat -f -e  --disable-subsumption-test --show-realization-mrses ace_mrs%(session)s' % {'grammar_dir':grammar_dir, 'iso':iso, 'session':session, 'count':str(len(mrs_files))}
```
goodmami, line 187: It looks like `count` is not used in the command.

goodmami (commented):
Hi, this is great to see, but it's maybe unfortunate that you've already put in so much effort in understanding ACE's [stdout protocols](https://pydelphin.readthedocs.io/en/latest/api/delphin.ace.html#ace-stdout-protocols) given that PyDelphin does this already. May I suggest using PyDelphin for the system calls to ACE?

Compiling ([ace.compile()](https://pydelphin.readthedocs.io/en/latest/api/delphin.ace.html#delphin.ace.compile)):

```pycon
>>> from delphin import ace
>>> ace.compile('path/to/config.tdl', 'grm.dat')
```

Generating ([ace.generate()](https://pydelphin.readthedocs.io/en/latest/api/delphin.ace.html#delphin.ace.generate); example using the ERG):

```pycon
>>> grm = '../erg-2018.dat'
>>> mrs = '[ TOP: h0 INDEX: e2 [ e TENSE: pres ] RELS: < [ _rain_v_1 LBL: h1 ARG0: e2 ] > HCONS: < h0 qeq h1 > ]'
>>> response = ace.generate(grm, mrs)
>>> [result['surface'] for result in response.results()]
['It rains.']
```

The PyDelphin documentation has a [guide](https://pydelphin.readthedocs.io/en/latest/guides/ace.html) and [API docs](https://pydelphin.readthedocs.io/en/latest/api/delphin.ace.html) to explain further.

Also, PyDelphin is already a dependency of the Grammar Matrix, so it should be already installed in your testing environment. Let me know if you have any questions.

I also have some specific comments in the code, in case you don't use PyDelphin, but I only looked at the part where it calls out to ACE and didn't review the rest.

gmcs/generate.py
rosypen: Hi Michael!

Thanks so much for your help and advice! I've made a new commit that now uses PyDelphin to call ace for both compiling and generating!

The commit is here - https://github.com/rosypen/matrix/commit/be5b86537e9cfdb72628411149d8297d9622beb9

The good news is it did clean up the code considerably and I got the parse trees, yay!

The bad news is I no longer have nicely formatted mrs because I don't think the tsdb output format does that for me. Is there a library I could use in PyDelphin for formatting the mrs?

Screen Shot 2022-08-13 at 1 55 20 AM

gmcs/generate.py
rosypen: It is not, I'm going to be removing it and the generate command as a whole BUT I will be bringing back the count later as a feature to specify max sentences to generate.

gmcs/generate.py
goodmami: > Is there a library I could use in PyDelphin for formatting the mrs?

Yes, there are options. See the list of [codecs](https://pydelphin.readthedocs.io/en/latest/api/delphin.codecs.html). If you just want SimpleMRS with indentation, that's easy:

```pycon
>>> from delphin.codecs import simplemrs
>>> m = simplemrs.decode('[ TOP: h0 INDEX: e2 [ e TENSE: pres ] RELS: < [ _rain_v_1 LBL: h1 ARG0: e2 ] > HCONS: < h0 qeq h1 > ]')
>>> print(simplemrs.encode(m, indent=True))
[ TOP: h0
  INDEX: e2 [ e TENSE: pres ]
  RELS: < [ _rain_v_1 LBL: h1 ARG0: e2 ] >
  HCONS: < h0 qeq h1 > ]
```

If you want to display DMRS or EDS instead, you'll first need to [convert the MRS](https://pydelphin.readthedocs.io/en/latest/guides/walkthrough.html#converting-semantic-representations).

And if you want a graphical web display, you can try [delphin-viz](https://github.com/delph-in/delphin-viz/).

gmcs/generate.py
goodmami: Ok, you can pass `"-n"` and the count as an command argument. But note that this is how many results to generate per input. E.g., if you pass in 3 MRSs with `"-n1"`, ACE will generate up to 3 sentences. I mention this because `str(len(mrs_files))` looks like you're counting how many input MRSs there are.

gmcs/generate.py
rosypen: Thank you again!

DMRS and EDS are pending TODOs that I may get to in the future. For now this has helped me get the simple mrs formatting!

Screen Shot 2022-08-13 at 3 06 36 PM

gmcs/generate.py
rosypen: Yes thank you for the clarification.

The old code I had there was just for me to test how the count option works and I put `str(len(mrs_files))` in there in some sense as a placeholder while I was figuring it out.

In the next few commits I will be adding an option for users to specify how many sentences they want to generate. Perhaps I can set a default max though.

PR #663, opened 2022-08-13
Using Pydelphin for Test by Generation

rosypen:
These changes reduce the code complexity considerably thanks to @goodmami for the pointers on how to use Pydelphin to call ACE for generation.

We now have parse trees and well formatted MRS on display!

Screen Shot 2022-08-13 at 3 06 36 PM

Up Next:
- Parsing the Test Sentences in the choices file and generating from those MRS
- Specifying max number of sentences to generate

The above two tasks will involve changing the Test by Generation Options page.

### DIFF
--- a/gmcs/generate.py
+++ b/gmcs/generate.py
@@ -9,6 +9,7 @@ import shutil
 import subprocess
 import json
 from delphin import ace
+from delphin.codecs import simplemrs
 
 display_gen_results_fn = '''
 (defun TbG-gen-results nil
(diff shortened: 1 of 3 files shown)


rosypen:
Oh I made another commit which I didn't realize would be automatically added to this pull request.

The new commit enables the feature to generate sentences using the test sentences in the choices file
- [Enable Generate from Test Sentences](https://github.com/rosypen/matrix/commit/35da715afafadff0c75f98a6aa5efd543a75785b)

Please click on the link above to see the full workflow of how that works. But In summary, we have this now:

Screen Shot 2022-08-15 at 12 14 57 AM

Screen Shot 2022-08-14 at 11 33 47 PM

emilymbender:
I'm happy to merge these, but I'd love some advice from those more familiar with github ... if I just click "merge pull request" on the latest, does it all get merged? Is there anything I need to be careful about?

arademaker:
I would suggest the `Squash and Merge` option so all intermediary commits in the source branch will be collapsed into a single commit in the trunk branch. 

image

arademaker:
Sorry, ignore the link I added in the last message; I just edited the message to fix it; I was trying to copy a screenshot.

arademaker:
It is an excellent practice always to have an issue associated with a PR, so the issue document the discussion that eventually was solved with a PR.

goodmami:
@emilymbender quick answer: yes, if you merge it will merge all the commits in this PR to the target branch (which is `trunk` here). If the target branch has changed since the PR was created and has introduced conflicts, GitHub would prevent you from merging until the conflicts were addressed.

I agree with @arademaker's suggestion to always have an issue for a PR. The issue describes the problem, and the PR describes the solution. When you close the issue via the PR, they get linked, and they both become a historical record of the development.

I don't recommend "Squash and merge" until you feel comfortable with Git and maybe not even then. It's a little more risky because it changes the history (local to the PR) which can introduce hard-to-fix conflicts. The following article lays out some pros/cons of the merge options: https://www.lloydatkinson.net/posts/2022/should-you-squash-merge-or-merge-commit/

emilymbender:
Thank you all --- I have merged this!

Issue #671, opened 2023-01-31
Lexical Multiple Inheritance Broken


Ubadub:
I believe [this PR](https://github.com/delph-in/matrix/pull/647) introduced a bug in the validate script that appears when specifying multiple supertypes on a noun I also believe I have a fix and would like to run it by @ltxom (the committer of that PR), @emilymbender + anyone else working on this.

## Bug description

Attempting to save a questionnaire (or upload a choices file) with a lexical type that has multiple supertypes will result in this error message:

<img width="1048" alt="Screenshot 2023-01-30 at 1 52 37 PM" src="https://user-images.githubusercontent.com/1286898/215635541-fad64965-fe0b-4e0b-929f-719cc7d7c4e4.png">

which has this Python stack trace:

<img width="964" alt="Screenshot 2023-01-30 at 1 53 44 PM" src="https://user-images.githubusercontent.com/1286898/215635525-1d4cadb2-4b89-4725-8095-1e83ed6ea56a.png">

## Steps to reproduce

A choices file that will reproduce this bug can be found here: [choices_supertype_bugreport.txt](https://github.com/delph-in/matrix/files/10541561/choices_supertype_bugreport.txt)

Alternatively, it can be reproduced with any choices file that has multiple inheritance for a lexical type, or on the website itself, by selecting multiple supertypes from the supertype dropdown selection box.

## Suspected Cause

[PR 647](https://github.com/delph-in/matrix/pull/647) introduced the following function to `utils.py` (direct [link](https://github.com/delph-in/matrix/blob/d74a5be38b0e1d781b015158ccbadeff8ce56875/gmcs/utils.py#L132)):

```
def recursive_get_supertypes_features(ch, lt, parent_features_list):
    feats = ch.get(lt).get('feat')
    for feat in feats:
        parent_features_list.append(feat)
    lt_supertypes = ch.get(lt).get('supertypes')
    if len(lt_supertypes) != 0:
        recursive_get_supertypes_features(ch, lt_supertypes, parent_features_list)
```

The bug occurs because `ch.get(lt).get('supertypes')` returns a string consisting of a comma-separated list of supertypes, for any lexical type with multiple supertypes. Presently, in all other places where similar functionality is required in the repo, there is always a `.split(', ')` call after, and then the resulting list is iterated over, whereas the code above assumes that there is only one supertype.

[...]

PR #673, opened 2023-02-01
Fixes delph-in#671 - bugfix for multiple inheritance for lexical types

Ubadub:
See discussion in #671 

@ltxom let me know if anything is not satisfactory or if the regression tests run into issues. Thank you!

### DIFF
--- a/gmcs/linglib/lexicon.py
+++ b/gmcs/linglib/lexicon.py
@@ -2,7 +2,7 @@ from collections import defaultdict
 
 from gmcs.linglib import case
 #from gmcs.linglib import lexical_items
-from gmcs.utils import get_name, recursive_get_supertypes_features
+from gmcs.utils import get_name
 from gmcs.choices import ChoiceDict
 from gmcs.linglib.lexbase import LexicalType, PositionClass
 from gmcs.linglib.lexbase import ALL_LEX_TYPES
@@ -178,6 +178,23 @@ def get_all_supertypes(key, choices):
                     supertypes.extend(parents)
     return supertypes, pathToRoot
 
+
+
+def get_all_supertypes_features(ch, lt):
+    """
+    Return a list of all features of all the supertypes of a given lexical type as defined by
+    the user. This does not include matrix internal types. Returns an empty list if no supertypes
+    defined.
+    """
+    lt_supertypes, pathToRoot = get_all_supertypes(lt, ch)
+    supertype_feats = []
+    for supertype in lt_supertypes:
+        if supertype:
+            supertype_dict = ch.get(supertype, '')
+            if supertype_dict:
+                for feat in supertype_dict.get('feat', ''):
+                    supertype_feats.append(feat)
+    return supertype_feats
 
 def get_lt_name(key, choices):
     if key in LEXICAL_SUPERTYPES:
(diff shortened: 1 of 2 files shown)


gmcs/linglib/lexicon.py
```diff
@@ -1080,22 +1097,16 @@ def validate_lexicon(ch, vr):
...
-                                       + lextype + ' class. If the value specified here is conflict with that other' \
-                                              ' specification, the grammar will not compile.'
-                                vr.warn(feat.full_key+'_name', mess)
+    # Abhinav 2023-01-31: Fixed bug blocking multiple inheritance, no longer recursive
+    # so no need to check for cycles (see discussion in issue #671)
+    for lextype in ALL_LEX_TYPES:
+        for lt in ch.get(lextype):
+            feat_name2full_key = {feat.get('name') : feat.full_key for feat in lt.get('feat') if feat}
+            supertype_feat_names = { feat.get('name', '') for feat in get_all_supertypes_features(ch, lt) }
+            dupe_feats = supertype_feat_names.intersection(feat_name2full_key.keys())
```
dantiston:
```suggestion
            feat_name2full_key = {feat.get('name'): feat.full_key for feat in lt.get('feat') if feat}
            supertype_feat_names = {feat.get('name', '') for feat in get_all_supertypes_features(ch, lt)}
            dupe_feats = supertype_feat_names.intersection(feat_name2full_key)
```

1. `dict`s iterate through their keys, so `supertype_feat_names.intersection(feat_name2full_key.keys())` is redundant to `supertype_feat_names.intersection(feat_name2full_key)`.
2. Styling: there shouldn't be a space before the colon on line 1104 or before or after the curly braces on 1105.

Ubadub: Thanks! Yes, I had realized that about the dictionary keys and meant to change it, but forgot to before committing/PRing this. Good catch!

ltxom:
Thank you. I verified these changes, which don't break other parts of the existing regression tests, and it solved the errors when validating  [choices_supertype_bugreport.txt](https://github.com/delph-in/matrix/files/10541561/choices_supertype_bugreport.txt).

@emilymbender Since it is a fix relating to the validation process, should we introduce a new regression test for it?

ltxom:
Emily mentioned a unit test could be added to verify validation improvements. The section about how to run/add a unit test is [here](https://github.com/delph-in/docs/wiki/MatrixCustomizationUnitTesting).

However, I followed the documentation above and found the parameter of "unit-test" is missing in the current Matrix trunk. I checked the commit history and found [Dr. Goodman updated it](https://github.com/delph-in/matrix/commit/dcb679a58569bd213e4e15d5a3b1dea8a3dbc9fe) to pytest.

> Tests can be run with `pytest`. For now this requires setting the
> PYTHONPATH variable:
> 
>     PYTHONPATH=. pytest tests/unit/choices_test.py
>     PYTHONPATH=. pytest tests/unit/validate_test.py
> 
> Don't just run all tests under `tests/unit/` because some are still
> broken. The goal is to rework these tests under a more standard pytest
> workflow so simply running `pytest` will be enough.

@Ubadub Do you have experience using pytest and adding a unit test?

Ubadub:
I tried running unit tests a couple days ago but ran into precisely the issue you have clarified above, so thank you for that. Yes, I've used pytest before though not in a few years but I'm sure I can figure it out. However, due to the demands of school it might be a few days before I can get around to it. I will keep you updated!

Ubadub:
Also, as I mentioned in the Issue thread- there actually are regression tests for this, but for customization only, it seems nothing in the regression tests workflow tests the validation script (except indirectly). So maybe we want to change that too? I will write the unit test as soon as I can but if you want me to look into adding a few lines for having regression testing check validate.py as well, I can do that.

emilymbender:
I'm confused as to the goal of this fix. It should be possible to do multiple inheritance on that page, I thought.

Ubadub:
It isn't possible, specifically for lexical types. I've attached a choices file with issue #671 that will allow you to reproduce this issue.

Issue #589, opened 2021-02-19
rename "my-head-adj-int" phrase


olzama:
The rules are added by adverbs_adpositions.py. The name of the phrase should be something distinct from just head-adj-int, under the current analysis, but it should of course not be "my".

emilymbender:
I just turned this one up, too. Also, this rule doesn't seem well integrated with other uses of head-adj. When adnominal possessives as modifiers are in the same grammar, ambiguity ensues. Who added this rule, and what is meant to be special about it? (The constraints added on the rule types seem redundant with the info that should be in the modifiers themselves, so I'm wondering what purposes they serve.)

olzama:
I'm afraid I added it, and I know there was a reason but I no longer remember because it's been a few years. It is likely that this was a confused analysis on my part (what you say about the modifiers). The easiest way to check would be to change the references (in the customization system) to head-adj and see what breaks in the wh- regression tests.

emilymbender:
@KerenR3 if you are interested in exploring more in this space, would you be up for trying the experiment that @olzama  suggests in this thread? That is -- find all of the places in the code where there is my-head-adj and change those to head-adj (look for my-adj-head, too, just in case) and run the regression tests and then report back here on what breaks.

KerenR3:
Sure, I can try that and see what happens.

emilymbender:
Thank you!

KerenR3:
It looks like doing the above causes two regression tests to fail (wh18-aux-inv-both and wh-dev-ckt-mini), both due to having extra parses.

Wh18-aux-inv-both fails because MOD is unconstrained on the subj-v-inv-rule (in the yes_no_questions library) which allows auxiliaries to serve as modifiers.

Wh-dev-ckt-mini fails because MOD is unconstrained on the specifier version of possessor_pron_lex (in the adnominal_possession library), allowing these possessor pronouns to act as modifiers.

emilymbender:
It's mysterious to me why the extra head-adj rules would get around this, but it also sounds like those are issues that should be fixed anyway. Can you try adding [ MOD < > ] to both of those other places and see how that fares on the regression tests?

KerenR3:
Adding MOD < > to the appropriate places causes the failing regression tests to pass without breaking any of the other tests. If I am understanding it correctly, the extra head-adj rules constrain the non-head daughter to be an adjective, adverb, or adposition which prevents either auxiliaries or possessor nouns from serving as modifiers.

[...]

PR #683, opened 2023-04-17
Fix for bug 672

KerenR3:
Added missing SYNSEM to the head-adj-int-phrase rule in wh_ques.py

### DIFF
--- a/gmcs/linglib/wh_ques.py
+++ b/gmcs/linglib/wh_ques.py
@@ -235,7 +235,7 @@ def customize_wh_ques(mylang, ch, rules, roots):
         # Rule out structural ambiguity for sentences like "Who sleeps where?"
         if ch.get('word-order') in ['svo', 'sov', 'osv']:
             mylang.add('''head-adj-int-phrase :+ [ 
-             LOCAL.CAT.VAL [ SUBJ clist, COMPS clist ] ] ].''', section='addenda')
+             SYNSEM.LOCAL.CAT.VAL [ SUBJ clist, COMPS clist ] ] ].''', section='addenda')
         if ch.get('word-order') == 'free':
             mylang.add(
                 '''adj-head-int-phrase :+ [ HEAD-DTR.SYNSEM.LOCAL.CAT.VAL.SUBJ clist ].''', merge=True)


dantiston:
Does this need an updated regression test? Since clearly there wasn’t a test before.

KerenR3:
I think there is an existing regression test that should have caught this problem, but the regression test passes even though  bug 672 leads to a grammar that does not load in the lkb. I think this might be due to the fact that there are some redundant lines for the head-adj related rules across adverbs_adpositions.py and wh_ques.py. Getting rid of the redundant lines in adverbs_adpositions.py causes the regression test to fail unless bug 672 is fixed.

emilymbender:
This doesn't make sense to me. If the grammar doesn't load, the regression test should be failing --- unless there's a larger bug in the regression testing system that counts tests as "passed" when there is no grammar.

Also, in what sense is the deleted material redundant? What error does the LKB give when trying to load such a grammar?

emilymbender:
More broadly, I'd love to resolve whatever is behind requiring a separate set of head-adj rules (my-head-adj-phrase, etc). I think these have to do with @olzama 's wh questions library. Olga, do you remember what the motivation was for creating separate types? (At the very least, the name of the types should reflect that motivation, but I suspect we should be able to collapse them into the main types.)

olzama:
I don't remember specifically. All I remember is I couldn't figure out how to use the existing types (I suppose trying to use them led to either compilation or regression test failures). But it was too long ago... This is the only mention that I found but it doesn't shed any light I'm afraid: https://github.com/delph-in/matrix/issues/589

dantiston:
> unless there's a larger bug in the regression testing system that counts tests as "passed" when there is no grammar.

I think the issue is that since the addition of  ICONS, the regression testing system relies solely on ACE determining if a grammar compiles, loads, and parses.

emilymbender:
That doesn't quite explain it though --- if the error that leads to the grammar not loading in the LKB also leads to the test passing, then there's more to it. Perhaps ace won't compile it either, and we're counting that as a "pass"?

KerenR3:
> This doesn't make sense to me. If the grammar doesn't load, the regression test should be failing --- unless there's a larger bug in the regression testing system that counts tests as "passed" when there is no grammar.
> 
> Also, in what sense is the deleted material redundant? What error does the LKB give when trying to load such a grammar?

The error in LKB states that LOCAL is introduced at multiple types for HEAD-ADJ-INT-PHRASE. The line I deleted (within adverbs_adpositions.py) is a feature constraint (VAL [ SUBJ clist, COMPS clist ] ]) on my-head-adj-phrase. my-head-adj-phrase inherits from head-adj-int-phrase, and head-adj-int-phrase has this same feature constraint specified as an amendment to the matrix.py version within wh_ques.py. 
For the regression test itself, the test logs at least show that sentences are being parsed, even when the grammar won't compile in the LKB.

emilymbender:
Hmm - what happens if you compile the grammar with ace? (Also, if you can, it would be great to have these grammars available to look at during Matrix/AGG in a few minutes.)

emilymbender:
Just a few notes from our discussion today: 
1) ace seems to be happily compiling a grammar with LOCAL introduced on two different types. Will investigate further.
2) that grammar passes the regression test, presumably because the extraneous constraint is ineffective (being in the wrong place in the tfs)

PR #704, opened 2024-02-06
Fix trans and intrans cases getting stuck in verification

Sam647254:
This issue occurs when a grammar has both required case-marking adpositions and complementizers. The `trans` case would be inserted multiple times into the `cases` list, which causes verification to fail on the customization pages.

### DIFF
--- a/gmcs/linglib/lexicon.py
+++ b/gmcs/linglib/lexicon.py
@@ -1013,16 +1013,17 @@ def validate_lexicon(ch, vr):
     for p in ch.patterns():
         p = p[0].split(',')
         c = p[0].split('-')
-        cases += [c[0], c[1]] if len(c) > 1 else [c[0]]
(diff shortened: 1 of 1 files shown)


Diana-BZ:
Hi Sam! Do you have a choices file I can use to test these changes?

Diana-BZ:
Hi Sam! Do you have an example screenshot of the options/inputs that are causing the verification issue?

Sam647254:
Sure, here is a screenshot of the error message that I'm seeing:
image
 
The options that cause this error span several pages, so it's probably easier if I share the choices file that can reproduce the error: 
[choices.txt](https://github.com/delph-in/matrix/files/14202074/choices.txt)

Running `matrix.py validate` on this file should also show that error, and after my fix is patched in, it should no longer appear.

Issue #653, opened 2022-07-22
Misplaced types in matrix.tdl


goodmami:
The following types are under the section for "Affixation" but seemingly do not belong there:

* `rule`
* `tree-node-label`
* `label`
* `meta`

These could be moved to a different (or new) section for improved understandability.

goodmami:
It's worth noting that all of these, except for `label` are not used anywhere.

Issue #634, opened 2022-02-24
basic-head-opt-subj not instantiated


emilymbender:
The attached choices file leads to a grammar that should have basic-head-opt-subj instantiated in rules.tdl, but doesn't
[choices.txt](https://github.com/delph-in/matrix/files/8129079/choices.txt)
.

Issue #617, opened 2021-10-04
Other Features questionnaire says FORM is only for verbs


olzama:
...which confuses people: https://delphinqa.ling.washington.edu/t/what-is-the-complementizer-form-for/677/3

Should probably change the wording. FORM can in practice (as well as in theory) be used for other things such as adpositions or complementizers.

PR #739, opened 2024-09-27
634 basic head opt subj

ajcassell:
#653 removed the three unused types and moved label into a new section called Trees
#617 small text change in questionnaire
#634 Added the joint conditional for subj-drop=subj-drop-all AND subj-con=subj-con-some. This fixed the bug and did not break any regression tests. 
Screen Shot 2024-09-27 at 2 42 13 PM


emilymbender:
Is the addition of basic-head-opt-subj adding any ambiguity anymore? (Also, I wonder if it's actually decl-head-opt-subj that should be added, despite what the bug report said.)

ajcassell:
> Is the addition of basic-head-opt-subj adding any ambiguity anymore? (Also, I wonder if it's actually decl-head-opt-subj that should be added, despite what the bug report said.)

It is not adding any ambiguity anymore. Originally I had added just

`if ch.get('subj-drop') == 'subj-drop-all':
        rules.add('basic-head-opt-subj := basic-head-opt-subj-phrase')` 

which is where the ambiguity came from because a few lines earlier in the code, the decl-head-opt-subj rule was also being instantiated:

`if ch.get('subj-drop') == 'subj-drop-all' and not (ch.get('subj-con') == 'subj-con-some'):
        rules.add('decl-head-opt-subj := decl-head-opt-subj-phrase.')
    if ch.get('subj-drop') == 'subj-drop-lex' and not (ch.get('subj-con') == 'subj-con-some'):
        rules.add('decl-head-opt-subj := decl-head-opt-subj-phrase.')`

So I decided to only add the basic-head-opt-subj rule as I did originally (when subj-drop=subj-drop-lex) but only when the specific criteria for decl-head-opt-subj was not met (subj-con=subj-con-some). This fixed the bug in that it instantiated the rule but also didn't break any of the regression tests because it no longer added the ambiguity of which rule to use. Does that seem right?

emilymbender:
Still not quite, actually. I think what's needed is decl-head-opt-subj in the case represented by that choices file which wasn't otherwise already giving decl-head-opt-subj.

Issue #412, opened 2020-05-19
Test by Generation PPP


goodmami:
**Migrated from Trac:**

> - Original: <http://lemur.ling.washington.edu/trac/matrix/ticket/700>
> - Reporter: waxd
> - Created: 08/23/11 16:33:35
> - Updated: 08/23/11 16:33:35
> - Keywords: 
> - Notes:
>   - There may be formatting problems from the conversion
>   - Comments on tickets are *not* migrated! See the original issue.

When using a choices file without an intransitive verb, clicking on the test by generation button throws a divide by zero error.

A choices file which causes this error is attached.

PR #742, opened 2024-10-02
Checking for length of vars before modulo

ajcassell:
#412 I was able to resolve this problem by adding length safeguard checks before the modulus operation. This allows for test by generation to work properly with choices files that either do not have transitive verbs or do not have intransitive verbs (which was the original issue). However, ace outputs an error file (below) indicating that that type of verb is missing. Is this alright?

Screen Shot 2024-09-30 at 5 12 34 PM
Screen Shot 2024-09-30 at 5 16 02 PM


emilymbender:
Yes, this looks fine.

Issue #374, opened 2020-05-19
doc strings


goodmami:
**Migrated from Trac:**

> - Original: <http://lemur.ling.washington.edu/trac/matrix/ticket/2>
> - Reporter: ebender
> - Created: 10/02/08 11:42:02
> - Updated: 10/28/11 16:05:24
> - Keywords: 
> - Notes:
>   - There may be formatting problems from the conversion
>   - Comments on tickets are *not* migrated! See the original issue.

Change comments before functions in .py files to doc strings.

PR #751, opened 2024-10-13
374 docstrings

ajcassell:
#374 

Curious whether it would be beneficial to add the arguments for each method that takes any. That appears as a standard in the docstrings documentation but I don't want to spend time doing that if you think it's not meaningful.


emilymbender:
As for adding info about methods --- I think that's a question best directed to @taraw28 @KerenR3 and others who have recently worked on this code. Would that have been useful to you?

Before I merge this one, what impact would that have on the currently failing regression tests?

PR #753, opened 2024-10-19
adding logging and proper exception raising

ajcassell:
I don't see any case where these exceptions could be thrown because the choices file is treated all as strings, but given that it was written this way, I left them and added logging and exception throwing. I made two changes to the regression tests which I believe were invalid text in their choices files. These caused the new exceptions to be thrown.

### DIFF
--- a/gmcs/choices.py
+++ b/gmcs/choices.py
@@ -410,9 +410,9 @@ class ChoicesFile:
                 self.load_choices(lines)
                 if type(choices_file) == str:
                     f.close()
-            except IOError:
-                print('Error loading choices from %s' % choices_file)
-                raise
+            except IOError as io:
+                message = 'Error loading choices from ' + choices_file + ": "
+                raise IOError(message + repr(io))
 
     def __str__(self):
         return str(self.choices)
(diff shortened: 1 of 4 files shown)


emilymbender:
Where do these messages go?

ajcassell:
I have had the same question about different logging styles throughout the code, so I decided to follow the other logs in this file but was not able to reproduce the logs. I have now changed this PR to raise the error with the custom log appended to the string of the original exception and was able to test this. The exception will be shown on the customization site or in the logs of a regression test.

Screen Shot 2024-10-22 at 10 29 28 AM
Screen Shot 2024-10-22 at 10 28 33 AM

emilymbender:
So on the customization site, we don't want people hitting the "purple page of pathos" but rather that first "error" page with the option to click through. Is that what happens here?

ajcassell:
If you are referring to this page, then yes that is what pops up first! I only included a screenshot above of what the error would tell the user. 

Screen Shot 2024-10-22 at 5 20 47 PM

emilymbender:
Great -- like that!

Issue #415, opened 2020-05-19
Don't autofill predicate for determiners


goodmami:
**Migrated from Trac:**

> - Original: <http://lemur.ling.washington.edu/trac/matrix/ticket/709>
> - Reporter: goodmami
> - Created: 11/16/11 19:54:57
> - Updated: 11/16/11 19:54:57
> - Keywords: 
> - Notes:
>   - There may be formatting problems from the conversion
>   - Comments on tickets are *not* migrated! See the original issue.

Many lexical items have their predicate auto-filled based on the orthography, but this shouldn't apply to determiners since we have things like def_q_rel, etc. Instead we can offer suggestions.

ajcassell:
@emilymbender what would be a good suggestion to include as text on the customization page if we want to not auto-fill the determiner predicates?

emilymbender:
I think the basics are exist_q_rel for existentials and all_q_rel for universals.

PR #754, opened 2024-10-19
warn validation on det preds + suggestion, tests

ajcassell:
#415 Don't autofill det preds, add suggestion, validation and tests

I made this validation a warn since all predicates are technically (as of right now) not validated for form so I wasn't sure whether blocking a user from making a freeform pred was a good idea. 

Screen Shot 2024-10-19 at 3 06 50 PM
Screen Shot 2024-10-19 at 3 06 41 PM

### DIFF
--- a/gmcs/linglib/lexicon.py
+++ b/gmcs/linglib/lexicon.py
@@ -1010,8 +1010,8 @@ def validate_lexicon(ch, vr):
                 mess = 'You must specify a predicate for each determiner you define.'
                 vr.err(stem.full_key + '_pred', mess)
 
-            if re.search("^_.*_q_rel$", stem.get('pred')) == None:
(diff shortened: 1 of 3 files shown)


emilymbender:
Since exist_q_rel and all_q_rel are grammar preds, they should not have the leading underscore.

emilymbender:
And the idea to make this a warning rather than an error makes sense.

ajcassell:
Ok, I have updated to not have the leading underscore.

emilymbender:
I think the regex should allow for both leading underscore & not. But the suggested form for exist_q_rel and all_q_rel shouldn't have one.

ajcassell:
Got it - I take it you merged this / the code is alright because the updated regex using .* will handle _.* and .*?

emilymbender:
Correct.

Issue #564, opened 2021-01-26
unary-poss-phrase-pron needs COORD - on daughter


emilymbender:
I'm seeing spurious ambiguity in a grammar with the specifications below, because the unary-poss-phrase-pron rules are allowing [ COORD + ] constituents as daughters (i.e. NP-B):

  cs9_np=on
  cs9_pat=a
    cs9_csap1_pat=fr1
    cs9_csap1_target=all

section=adnom-poss
  poss-pron1_type=non-affix
  poss-pron1_agr=non-agree
  poss-pron1_mod-spec=spec
  poss-pron1_order=head-final
    poss-pron1_instance1_name=1st-sg-possessive
    poss-pron1_instance1_orth=in
      poss-pron1_instance1_feat1_name=person
      poss-pron1_instance1_feat1_value=1st
      poss-pron1_instance1_feat2_name=number
      poss-pron1_instance1_feat2_value=sg
    poss-pron1_instance2_name=1st-pl-possessive
    poss-pron1_instance2_orth=itom
      poss-pron1_instance2_feat1_name=number
      poss-pron1_instance2_feat1_value=pl
      poss-pron1_instance2_feat2_name=person
      poss-pron1_instance2_feat2_value=1st
    poss-pron1_instance3_name=2nd-sg-possessive
    poss-pron1_instance3_orth=em=
    poss-pron1_instance4_name=2nd-pl-possessive
    poss-pron1_instance4_orth=enchim
      poss-pron1_instance4_feat1_name=number
      poss-pron1_instance4_feat1_value=pl
      poss-pron1_instance4_feat2_name=person
      poss-pron1_instance4_feat2_value=2nd
    poss-pron1_instance6_name=3rd-sg-possessive-alt
    poss-pron1_instance6_orth=apo'ik
      poss-pron1_instance6_feat1_name=number
      poss-pron1_instance6_feat1_value=sg
      poss-pron1_instance6_feat2_name=person
      poss-pron1_instance6_feat2_value=1st
    poss-pron1_instance7_name=3rd-pl-possessive
    poss-pron1_instance7_orth=vempo'im
      poss-pron1_instance7_feat1_name=number
      poss-pron1_instance7_feat1_value=pl
      poss-pron1_instance7_feat2_name=person
      poss-pron1_instance7_feat2_value=3rd
  poss-pron2_type=non-affix
  poss-pron2_agr=non-agree
  poss-pron2_mod-spec=spec
  poss-pron2_order=head-final
[...]

PR #760, opened 2024-11-01
COORD - on daughter, MOD <> on mother

ajcassell:
#564


emilymbender:
This looks like it's taking care of half of #564.

ajcassell:
> This looks like it's taking care of half of #564.

Which part is missing?

emilymbender:
The part about MOD.

ajcassell:
The MOD <> already exists on the mother. If I interpreted our slack thread correctly, isn't the MOD specified there on line 126? If you are referring to the poss-unary-pron phrase, I did not add the MOD because the issue only mentioned COORD - for that rule. I can add it there as well if needed though.

emilymbender:
"poss-unary-phrase has a similar issue, where it should say [ COORD - ] on the daughter and [ MOD < > ] on the mother."

ajcassell:
This is the poss-unary-phrase, right?
Screen Shot 2024-11-01 at 1 41 44 PM

emilymbender:
I assume so -- and it looks like it is there. So let's call it good.

Issue #577, opened 2021-02-10
Prompting for PRED values for wh pronouns


emilymbender:
Based on the experience in this year's 567, it's not obvious that something like _thing_n_rel is expected rather than _what_n_rel, in defining a wh pronoun.

ajcassell:
@emilymbender what would be a reasonable "prompt" here?

emilymbender:
Could be something like: "Use _thing_n_rel, _person_n_rel, _place_n_rel or similar"

PR #762, opened 2024-11-07
adding suggestion string

ajcassell:
#577 I tried to make this text suggestion only show up when the question pronoun box is selected but this doesn't seem possible in the realm of the matrixdef syntax

Screen Shot 2024-11-06 at 6 07 50 PM

### DIFF
--- a/gmcs/linglib/lexicon.py
+++ b/gmcs/linglib/lexicon.py
@@ -408,10 +408,21 @@ def validate_lexicon(ch, vr):
             vr.warn(n.full_key + '_det', mess % message_map[det])
             
         q = n.get('inter')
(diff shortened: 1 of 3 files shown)


emilymbender:
If that shows up under every pronoun, that would be suboptimal, I think. What about using validation instead, and showing this as a warning just in case someone enters a string matching one of what, who, where, when in that box?

ajcassell:
Alright, I've changed to a warning instead.
Screen Shot 2024-11-10 at 4 18 52 PM
Screen Shot 2024-11-10 at 4 19 10 PM
Screen Shot 2024-11-10 at 4 19 42 PM

Issue #599, opened 2021-03-19
Too many values in trigger rules


emilymbender:
Many of the choices files created for Ling 567 in 2021 gave trigger rules like this:

ki_3_gr := arg0e_gtr &
  [ CONTEXT.RELS.LIST < [ ARG0.E [ TENSE fut,
                                   ASPECT pot,
                                   MOOD oblig,
                                   cond  ] ] >,
    FLAGS.TRIGGER "ki_3" ].

... which lead to grammars that don't load.

emilymbender:
This seems to happen when the TAM features are specified on the auxiliary, rather than its complement. The attached choices file illustrates the behavior.
[choices.txt](https://github.com/delph-in/matrix/files/14489925/choices.txt)

ajcassell:
@emilymbender So looking at this choices file, this grammar wouldn't load in the LKB because of this issue that I recently put in validation for: https://github.com/delph-in/matrix/issues/464
When I make aux1_subj=np instead of aux1_subj=np-comp-case to pass validation, I am able to load the resulting grammar in the LKB. This makes me think this issue is not related to TAM features on the auxiliary. However, if TAM features on the aux rather than complement is a true issue, perhaps we could morph this issue into that?

emilymbender:
What do the generated trigger rules look like in the grammar that can load?

ajcassell:
I am actually seeing identical trigger.mtr files for the choices file above using the live site right now and the test site with my change to np-comp-case.
<img width="547" alt="Screen Shot 2024-11-12 at 4 44 39 PM" src="https://github.com/user-attachments/assets/de19345e-6f4a-451c-b3a7-501359d0f4d9">

emilymbender:
So that trigger rule is broken -- there should be one value for TENSE that is the shared supertype of present & future. If that supertype isn't already created, the production of the trigger rule should also involve creating it.

ajcassell:
I see. That supertype is not already created from what I see. The choices/customization site show past, present, and future as the three tenses. The reason they show up together in the trigger rule is because it's populated directly from the choices i.e. ```    aux1_feat1_name=tense
    aux1_feat1_value=present, future
    aux1_feat1_head=verb```
I don't know if there are any restrictions on which combination of tenses are allowed on a feature, but what should the supertype be? Can it be nonpast (also one of the options that is not selected on the TAM page)? Furthermore, would we need to add a supertype for each combination of tenses?

emilymbender:
[...]

PR #767, opened 2024-11-16
looking for supertype name for TAM features on aux to create correct trigger rule

ajcassell:
#599 
Trigger rule now looks like:
Screen Shot 2024-11-15 at 6 31 45 PM

Second portion of this is to fix something that came up from https://github.com/delph-in/matrix/pull/765
Although all the regression tests passed with that PR ^, I just noticed when trying to load a grammar into the LKB that there was an error loading because NON-HEAD-DTR was defined in two places (ternary-headed-phrase and basic-binary-headed-phrase) so I looked for where I could put it on a supertype instead of each. I'm not sure WHY it can't be defined twice, but once I changed it to this, all tests still pass and the LKB loads the grammar and parses. I see that there are other things that also inherit from headed-phrase so I'm not sure if it is okay to put there. Please let me know if this is incorrect and if there is a different solution.

### DIFF
--- a/gmcs/linglib/auxiliaries.py
+++ b/gmcs/linglib/auxiliaries.py
@@ -347,8 +347,7 @@ def get_users_type_name(aux):
     userstypename = name + '-aux-lex'
     return userstypename
 
-
-def add_auxiliaries_to_lexicon(userstypename, sem, aux, lexicon, trigger):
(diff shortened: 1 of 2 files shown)


emilymbender:
The LKB requires each feature to be declared on a single, most general type which is inherited by any type that mentions that feature. (I'm a little surprised that ace apparently doesn't actually!)

PR #770, opened 2024-11-24
adding namespace collision logic for sentential negation page with auxes in lexicon

### LINKED ISSUE #597: Negation bug: Lex entry with two supertypes
The attached choices file produces this broken lexical entry:

poen := neg-aux-lex & neg-comp-lex &
  [ SYNSEM.LKEYS.KEYREL.PRED "neg_rel",
    STEM < "poen" > ].

[buggy-neg.txt](https://github.com/delph-in/matrix/files/6174369/buggy-neg.txt)

ajcassell:
This is to fix #597 
I updated the choices file that was included in the issue since it did not pass more recent validation. Updated copy is attached. 
[buggy-neg-final.txt](https://github.com/user-attachments/files/17891593/buggy-neg-final.txt)
This will add numeric suffixes to the aux since this goes through the same namespace collision logic as nouns and verbs. Adverbs on the other hand will not have such collisions so remain as the originally named entry. The stem is the same but now there is a separate lexical entry for the adverb and the auxiliary instead of one entry thatt inherits from both rules. 
Screen Shot 2024-11-23 at 9 53 36 PM

### DIFF
--- a/gmcs/linglib/lexical_items.py
+++ b/gmcs/linglib/lexical_items.py
@@ -16,6 +16,8 @@ from gmcs.linglib.nominalized_clauses import needs_anc_wo_feat
 from gmcs.linglib.parameters import determine_vcluster
 from gmcs.linglib.lexbase import ALL_LEX_TYPES, LEXICAL_SUPERTYPES
 from gmcs.linglib.lexicon import get_all_supertypes
+from gmcs.linglib.negation import get_neg_stemids
+from gmcs.linglib.negation import add_neg_name
 from gmcs.linglib.clausalmods import get_subord_stemids
 from gmcs.linglib.clausalmods import add_subord_name
 from gmcs.feature_type_use import USED_TYPES
@@ -109,10 +111,11 @@ def insert_ids(ch):
                     stemids[id] += 1
                 else:
                     stemids[id] = 1
-    # KPH Subordinators are added outside the lexicon,
+    # KPH Subordinators and negators are added outside the lexicon,
     # but should still be checked for possible name-space collisions
     # this should be a temporary/hacky fix- a bug has been filed
     stemids = get_subord_stemids(ch, stemids)
+    stemids = get_neg_stemids(ch, stemids)
 
     # Now that stemids has the full count, go through and add
     # to the choices file object.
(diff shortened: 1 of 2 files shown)


emilymbender:
I think that numbers should be added to adverbs as well. The namespace collision is in the lexical identifier (the part before := in lexicon.tdl) and that is one giant namespace, not divided by lexical type.

ajcassell:
Yes, the namespace is the same but where the logic happens to look at namespace collisions only happens for nouns, verbs and auxiliaries because there can be multiple of them. If I remember correctly, there wasn't an easy way/place to look at all the names at once, but I will look again!

ajcassell:
After looking at the code in depth again, these are my findings (apologies if this is in the technical weeds):
We loop through part of speeches on lexicon page (which includes adv) to create ids for each orth, but because the negation adverb is not listed on the lexicon page, the orth is not added here. I followed the pattern that was used for clausal mods as well. That's the reason why the adverb gets added with original lexical identifier and then other entries with that POS and orth that conflict would get the number suffixes added. The adverb orth is added by looking at comp-neg-orth field in choices which is not as simple as finding the lexical entries orth values. This is just an example of one of the orths added from negation page but there are the 8 in the list in this PR. It would be odd/expensive(?) to search for these 8 different orth values in the choices object after creating the other lexical entries. An alternative I tried was to create any of the neg orths with '_1' to begin with, but this led to the original issue this PR is trying to solve. 
The correct entry does get added to the temp choices using the current naming logic which is great and what we would like: comp-neg-orth_name=poen_2, but since comp-neg-orth=poen is above and it's added in a separate area in the negation library, I realized the _name is used to create the lexicon entry so that's why it works for all of those on the lexicon page and not those added on the sentential negation page. I'm wondering if I could add a method in negation.py that adds negation entries to the lexicon AFTER all the id/name counting is handled. However, this would mean I have to take all 13 of the snippets like 
lexicon.add(TDLencode(orth) + ' := neg1-comp-lex &\
                [ STEM < \"' + orthstr + '\" >,\
                  SYNSEM.LKEYS.KEYREL.PRED \"neg_rel\" ].')
to this method and whatever conditional is present to trigger them. Should I try that?

emilymbender:
Got it. In that case, poen for the adverb and poen_1 for the other makes sense.  Is that the current state of affairs?

ajcassell:
Yes that is correct!

PR #771, opened 2024-11-24
652 unusedtypes

ajcassell:
https://github.com/delph-in/matrix/issues/653 removed the three unused types and moved label into a new section called Trees
https://github.com/delph-in/matrix/issues/617 small text change in questionnaire


emilymbender:
I think "node labels" is a clearer heading than "trees". Can you update that?

